PR #22, opened 2017-01-22
[HIVEMALL-30] Increase -Xmx to -Xmx1536m to avoid OutOfMemoryError

wangyum:

Increase -Xmx to -Xmx1536m to avoid `OutOfMemoryError`.

## How was this patch tested?

Manual tests

### DIFF
--- a/spark/spark-2.0/pom.xml
+++ b/spark/spark-2.0/pom.xml
@@ -33,7 +33,7 @@
 
 	<properties>
 		<PermGen>64m</PermGen>
-		<MaxPermGen>1024m</MaxPermGen>
+		<MaxPermGen>512m</MaxPermGen>
 		<CodeCacheSize>512m</CodeCacheSize>
 		<main.basedir>${project.parent.basedir}</main.basedir>
 	</properties>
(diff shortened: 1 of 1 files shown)


myui:
@maropu Could you take a look at?

maropu:
TravisCI always pass these tests though, why did you got the error? What's your environment?

maropu:
Oh, I got you. So, how about setting these values along with the spark one (e.g., -mx3g)? See: https://github.com/apache/spark/blob/master/pom.xml#L2066

maropu:
okya, LGTM cc: @myui

myui:
@maropu BTW, why `mvn -q scalastyle:check test -Pspark-2.0" exited with 1` is happening?

maropu:
@myui In running these tests, a spark context and unit tests work in the same JVM. The context uses at most 1g memory by default. So, in some conditions, the unit tests run short of memory, I think. If we set 3g at the value, the context uses 1g and the tests do 2g. That sounds fine to me.

myui:
@maropu It seems GC is still happening for some cases...
```
HivemallFeatureOpsSuite:
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

`MaxPermGen=1024m` might be too big to need more space for others. Do we need `-MaxPermGen` ? Less parameter is better for configuring JVM.

maropu:
@myui How about 512m? Sorry, but I forgot why I set that value there though, I checked the spark configure and I found that value is 512m.

wangyum:
I'll try to set 512m.

myui:
@wangyum @maropu Thanks. Merged with some modifications. Configuration for spark-1.6 should also be changed.

maropu:
okay, thanks!

PR #45, opened 2017-02-13
[HIVEMALL-71] Handle null values and add a unit Tests to RescaleUDF.

wangyum:

Handle null values and add a unit Tests to `RescaleUDF`.

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-71

## How was this patch tested?

unit tests

### DIFF
--- a/core/src/main/java/hivemall/ftvec/scaling/RescaleUDF.java
+++ b/core/src/main/java/hivemall/ftvec/scaling/RescaleUDF.java
@@ -62,9 +62,17 @@ public final class RescaleUDF extends UDF {
             throws HiveException, IllegalArgumentException {
         String[] fv = s.split(":");
         if (fv.length != 2) {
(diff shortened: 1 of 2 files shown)


myui:
Thanks, I will take this PR on.

myui:
@wangyum Thanks. Merged with some modifications.

```sql
select rescale(v,min,max) from (
    select cast(1.2 as float) as v, 1.0 as min, 10.0 as max
    union all
    select cast(null as float) as v, 1.0 as min, 10.0 as max
    union all
    select cast(1.2 as double) as v, 1.0 as min, 10.0 as max
    union all
    select cast(null as double) as v, 1.0 as min, 10.0 as max
    union all
    select cast(1.2 as double) as v, cast(5.0 as double) as min, cast(10.0 as float) as max
) t;

0.022222228
NULL
0.022222228
NULL
-0.76
```

PR #46, opened 2017-02-16
[HIVEMALL-72] Fix corner-case rescale UDF behavior to return range [0.0,1.0]

myui:

Fix rescale UDF behavior to return range `[0.0,1.0]`.

## What type of PR is it?

Bug Fix

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-72

## How was this patch tested?

manual tests

```sql
-- Before
select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 1.5999999	-0.65

-- After
select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 1.0 0.0
```

### DIFF
--- a/core/src/main/java/hivemall/ftvec/scaling/RescaleUDF.java
+++ b/core/src/main/java/hivemall/ftvec/scaling/RescaleUDF.java
@@ -107,10 +107,10 @@ public final class RescaleUDF extends UDF {
             return 0.5f;
         }
         if (value < min) {
-            return min;
+            return 0.f;
         }
         if (value > max) {
-            return max;
+            return 1.f;
         }
         return (value - min) / (max - min);
     }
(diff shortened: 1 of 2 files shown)


myui:
Oops.. the latter one is expected.

```sql
select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 3.0 1.0

select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 1.0 0.0
```

myui:
@wangyum Are you agree with the following new behavior of [rescale(value,min,max)](https://en.wikipedia.org/wiki/Feature_scaling#Rescaling)?

```
-- new
select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 1.0 0.0

-- old
select rescale(4.2,1.0,3.0),rescale(-0.3, 1.0, 3.0);
> 1.5999999	-0.65
```

Throwing an exception if value is out of range but exact min/max is sometimes unknown beforehand.

wangyum:
@myui I agree with you.

PR #59, opened 2017-03-04
[HIVEMALL-85] Upgrade hivemall-xgboost's hadoop version to 2.6.5

wangyum:

Upgrade hivemall-xgboost's hadoop version to 2.6.5

## What type of PR is it?

Improvement

## What is the Jira issue?

[HIVEMALL-85](https://issues.apache.org/jira/browse/HIVEMALL-85)

## How was this patch tested?

manual tests

### DIFF
--- a/xgboost/pom.xml
+++ b/xgboost/pom.xml
@@ -41,8 +41,8 @@
 		<!-- provided scope -->
 		<dependency>
 			<groupId>org.apache.hadoop</groupId>
-			<artifactId>hadoop-core</artifactId>
-			<version>0.20.2-cdh3u6</version>
+			<artifactId>hadoop-client</artifactId>
+			<version>2.6.5</version>
 			<scope>provided</scope>
 		</dependency>
 		<dependency>


maropu:
@wangyum Thanks for your continuous contributions! 
@myui do we have any reason to have a dependency with hadoop-core `0.20.2-cdh3u6`? I just used this dependency along with the other modules.

myui:
@wangyum @maropu I think it's `provided` maven space and thus no need to change.
`provided` scope libraries are used just for compiling and not bundled in jars.

maropu:
Yea, I know, but why we use cdh instead of hadoop community version?

myui:
Mostly for historical reasons. Better to change the version dependencies to 2.4.0 but dependencies packages change and careful verification is required.

This branch changed the dependencies to 2.4.0 
https://github.com/myui/hivemall/tree/dev/yarnkit

maropu:
@wangyum Why you select `2.6.5` in this pr? Any reason?

wangyum:
Be consistent with the Spark.
https://github.com/apache/spark/blob/d76c18746bdd0908bae89e62da9667001be75624/pom.xml#L125

myui:
We need to change whole dependencies to v2.4.0 following the way in https://github.com/myui/hivemall/tree/dev/yarnkit

File a ticket https://issues.apache.org/jira/browse/HIVEMALL-86

I would like to move the discussions to [Hivemall-86](https://issues.apache.org/jira/browse/HIVEMALL-86).

@wangyum Could you close this pull request?

wangyum:
OK

myui:
Thanks!

PR #61, opened 2017-03-08
[HIVEMALL-88][SPARK] Support a function to flatten nested schemas

maropu:
Since parquet is first-citizen in Spark, most of users use nested schemas in DF. In some usecases, they need to flatten nested schemas into flat ones to dump data into the other systems that do not support nested schemas, e.g., RDBMS. So, this pr added a conversion function from nested schemas to flat ones.

## What type of PR is it?
Improvement

## What is the Jira issue?
https://issues.apache.org/jira/browse/HIVEMALL-88

## How was this patch tested?
Added tests in `HivemallOpsSuite`.

### DIFF
--- a/docs/gitbook/SUMMARY.md
+++ b/docs/gitbook/SUMMARY.md
@@ -163,6 +163,7 @@
 
 * [Generic features](spark/misc/misc.md)
     * [Top-k Join processing](spark/misc/topk_join.md)
+    * [Ohter utility functions](spark/misc/functions.md)
 
 ## Part X - External References
 
--- a/docs/gitbook/SUMMARY.md
+++ b/docs/gitbook/SUMMARY.md
@@ -163,7 +163,7 @@
 
 * [Generic features](spark/misc/misc.md)
     * [Top-k Join processing](spark/misc/topk_join.md)
-    * [Ohter utility functions](spark/misc/functions.md)
+    * [Other utility functions](spark/misc/functions.md)
 
 ## Part X - External References
 
--- a/spark/spark-2.1/src/main/scala/org/apache/spark/sql/hive/HivemallOps.scala
+++ b/spark/spark-2.1/src/main/scala/org/apache/spark/sql/hive/HivemallOps.scala
@@ -805,13 +805,31 @@ final class HivemallOps(df: DataFrame) extends Logging {
     JoinTopK(kInt, df.logicalPlan, right.logicalPlan, Inner, Option(joinExprs.expr))(score.named)
   }
 
-  private def doFlatten(schema: StructType, prefix: Option[String] = None) : Seq[Column] = {
+  private def doFlatten(schema: StructType, separator: Char, prefixParts: Seq[String] = Seq.empty)
+    : Seq[Column] = {
     schema.fields.flatMap { f =>
-      val colName = prefix.map(p => s"$p.${f.name}").getOrElse(f.name)
+      val colNameParts = prefixParts :+ f.name
       f.dataType match {
-        case st: StructType => doFlatten(st, Option(colName))
-        case _ => col(colName).as(colName) :: Nil
+        case st: StructType =>
+          doFlatten(st, separator, colNameParts)
+        case _ =>
+          col(colNameParts.mkString(".")).as(colNameParts.mkString(separator.toString)) :: Nil
       }
+    }
+  }
+
+  // Converts string representation of a character to actual character
+  @throws[IllegalArgumentException]
+  private def toChar(str: String): Char = {
+    if (str.length == 1) {
+      str.charAt(0) match {
+        case '$' | '_' | '.' => str.charAt(0)
+        case _ => throw new IllegalArgumentException(
+          "Must use '$', '_', or '.' for separator, but got " + str)
+      }
+    } else {
+      throw new IllegalArgumentException(
+        s"Separator cannot be more than one character: $str")
     }
   }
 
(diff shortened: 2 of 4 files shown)


docs/gitbook/SUMMARY.md
```diff
@@ -163,6 +163,7 @@
 
 * [Generic features](spark/misc/misc.md)
     * [Top-k Join processing](spark/misc/topk_join.md)
+    * [Ohter utility functions](spark/misc/functions.md)
```
takuti: typo~

maropu: Fixed and Thanks!

spark/spark-2.1/src/main/scala/org/apache/spark/sql/hive/HivemallOps.scala
```diff
@@ -805,6 +805,47 @@ final class HivemallOps(df: DataFrame) extends Logging {
     JoinTopK(kInt, df.logicalPlan, right.logicalPlan, Inner, Option(joinExprs.expr))(score.named)
   }
 
+  private def doFlatten(schema: StructType, prefix: Option[String] = None) : Seq[Column] = {
+    schema.fields.flatMap { f =>
+      val colName = prefix.map(p => s"$p.${f.name}").getOrElse(f.name)
```
myui: Dot `.` is a special symbol in SQL. So, better to change the default separator (e.g., `$` ) and have an option to choose the separator. `news20.train` means train table in news20 database.

SQL identifiers are depending on RDBMSs but generally includes `[0-9,a-z,A-Z$_]`.
https://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html
https://msdn.microsoft.com/en-us/library/ms175874.aspx
https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

maropu: I know, but this is a Spark-local specification. So, the change you suggested make `doFlatten` fail.

myui: ??? flatten'ed column names are not a Spark specification. Better to have a ability to use a custom separator.

maropu: In Spark, the dot is used as the separator of column names in nested schema.
Currently, Spark users cannot change this separator via configurations.
For example,

```
scala> val ds = Seq((1, (1.0, "a"))).toDS()
ds: org.apache.spark.sql.Dataset[(Int, (Double, String))] = [_1: int, _2: struct<_1: double, _2: string>]

scala> ds.printSchema
root
 |-- _1: integer (nullable = false)
 |-- _2: struct (nullable = true)
 |    |-- _1: double (nullable = false)
 |    |-- _2: string (nullable = true)


scala> ds.select($"_2._2").show
+---+
| _2|
+---+
|  a|
+---+
```

maropu: So, the dot is more natural for Spark users.

maropu: Ah, I found an issue;
```
scala> val df = Seq((1, (1.0, "a"))).toDF()
df: org.apache.spark.sql.DataFrame = [_1: int, _2: struct<_1: double, _2: string>]

scala> val ds1 = df.flatten().select("_2._1")
org.apache.spark.sql.AnalysisException: cannot resolve '`_2._1`' given input columns: [_1, _2._1, _2._2];;
'Project ['_2._1]
+- Project [_1#67 AS _1#73, _2#68._1 AS _2._1#74, _2#68._2 AS _2._2#75]
   +- LocalRelation [_1#67, _2#68]

  at org.apache.spark.sql.catalyst.analysis.package$AnalysisErrorAt.failAnalysis(package.scala:42)
  at org.apache.spark.sql.catalyst.analysis.CheckAnalysis$$anonfun$checkAnalysis$1$$anonfun$apply$2.applyOrElse(CheckAnalysis.scala:75)
  at org.apache.spark.sql.catalyst.analysis.CheckAnalysis$$anonfun$checkAnalysis$1$$anonfun$apply$2.applyOrElse(CheckAnalysis.scala:72)
  at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$transformUp$1.apply(TreeNode.scala:289)
```
So, I'll reconsider this and please give me a sec. Thanks.

myui: `ds.select($"_2._2")` works for nested scheme. So, flatten'ed `ds.flatten` should be accessed without `.`.

maropu: Actually, we can access this column like this;
```
scala> val df = Seq((1, (1.0, "a"))).toDF()
df: org.apache.spark.sql.DataFrame = [_1: int, _2: struct<_1: double, _2: string>]

scala> val ds1 = df.flatten().select("`_2._1`").show
+-----+
|_2._1|
+-----+
|  1.0|
+-----+

```

maropu: @myui How about the latest fix? As you suggested, I added an option for separator.

myui: @maropu 👍 LGTM. Could you merge and close JIRA ticket as FIXED?

maropu: okay, thanks. I'll merge soon.

maropu:
Merged to master.

PR #63, opened 2017-03-18
[HIVEMALL-90] Refine incomplete AUC UDAF implementation

takuti:
## What changes were proposed in this pull request?

Since AUC UDAF (classification) did not work correctly for some specific merge orders, this PR fixes the issue by modifying the UDAF's `merge()` and `terminate()` implementation.

Moreover, unit tests are refined accordingly, and a utility method is created in **HiveUtils**.

## What type of PR is it?

Bug Fix

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-90

## How was this patch tested?

- Unit test
- Manual test on EMR

## How to use this feature?

Nothing has been changed from [current AUC UDAF](https://hivemall.incubator.apache.org/userguide/eval/auc.html).

PR #66, opened 2017-04-03
[HIVEMALL-91] Implement Online LDA

takuti:

Implement Online LDA model and its UDTF

## What type of PR is it?

Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-91

## How was this patch tested?

Unit test and manual test

## How to use this feature?

See [HERE](https://gist.github.com/takuti/d24324e76d4b2ec7dc4b1d50a4d192d8)


core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,497 @@
...
+            dummyLambdas[b] = tmpDummyLambda;
+        }
+    }
+
+    public void train(String[][] miniBatch, int time) {
+        miniBatchSize = miniBatch.length;
+
+        rhot = Math.pow(tau0_ + time, -kappa_);
+
+        if (printLambda) {
```
myui: please remove all debug prints. 

Use `Logger.debug()` for debug prints.

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,497 @@
...
+    private HashMap<String, float[]> lambda_;
+
+    // check convergence in the expectation (E) step
+    private double DELTA = 1E-5;
+
+    // for update size of lambda_
+    private int dummySize = 100;
+    private float[][] dummyLambdas;
+
+    private ArrayList<HashMap<String, Float>> miniBatchMap;
```
myui: ArrayList -> List
HashMap -> Map

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,497 @@
...
+    private void updateSizeOfParameterForMiniBatch() {
+        phi_ = new ArrayList<HashMap<String, float[]>>();
+        gamma_ = new float[miniBatchSize][];
+
+        for (int d = 0; d < miniBatchSize; d++) {
+            float[] gammad = getRandomGammaArray();
+            gamma_[d] = gammad;
+
+            // phi_ not needed to be initialized
+            HashMap<String, float[]> phid = new HashMap<String, float[]>();
```
myui: HashMap => Map

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,497 @@
...
+    public double getLambda(String label, int k) {
+        Preconditions.checkArgument(lambda_.containsKey(label),
+            "Word `" + label + "` is not in the corpus.");
+        Preconditions.checkArgument(k < lambda_.get(label).length,
+            "Topic index must be in [0, " + lambda_.get(label).length + "]");
+        return lambda_.get(label)[k];
+    }
+
+    public void showTopicWords() {
+        System.out.println("SHOW TOPIC WORDS:");
```
myui: Use Logger.info() or Logger.debug()

core/src/main/java/hivemall/lda/OnlineLDAUDTF.java
```diff
@@ -0,0 +1,112 @@
...
+		this._kappa = kappa;
+		this._iterations = iterations;
+		this._delta = delta;
+
+		return cl;
+	}
+	
+	@Override
+	public StructObjectInspector initialize(ObjectInspector[] argOIs) throws UDFArgumentException {
+		// TODO Auto-generated method stub
```
myui: not implemented yet.

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+        gd = new GammaDistribution(SHAPE, SCALE);
+        gd.reseedRandomGenerator(1001);
+
+        // initialize the parameters
+        lambda_ = new HashMap<String, float[]>();
+
+        setDummyLambda();
+    }
+
+    private void setDummyLambda() {
```
myui:
```java
@Nonnull
private final float[][] dummyLambdas;

this.dummyLambdas = dummyLambda();

@Nonnull
private static double[] dummyLambda() {
}
```

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+            float[] tmpDummyLambda = new float[K_];
+            tmpDArray = gd.sample(K_);
+            for (int k = 0; k < K_; k++) {
+                tmpDummyLambda[k] = (float) tmpDArray[k];
+            }
+            dummyLambdas[b] = tmpDummyLambda;
+        }
+    }
+
+    public void train(String[][] miniBatch, int time) {
```
myui:
```java
@Nonnull String[][] miniBatch
```

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+
+        // Maximization
+        stepM();
+    }
+
+    private void stepE() {
+        // 1) Updating phi_
+
+        // dirichlet_expectation_2d(gamma_)
+        float[][] ElogTheta = new float[miniBatchSize][K_];
```
myui: Change `eLogTheta` to camel case for local variables.

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+
+    // for update size of lambda_
+    private int dummySize = 100;
+    private float[][] dummyLambdas;
+
+    private List<Map<String, Float>> miniBatchMap;
+    private int miniBatchSize;
+
+    private int accumDocCount = 0;
+    private int accumWordCount = 0;
```
myui: consistent naming rule for class member variables.  
e.g., `_accumWordCount`.

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+    private double SHAPE = 100d;
+    private double SCALE = 1.d / SHAPE;
+
+    // parameters
+    private List<Map<String, float[]>> phi_;
+    private float[][] gamma_;
+    private Map<String, float[]> lambda_;
+
+    // check convergence in the expectation (E) step
+    private double DELTA = 1E-5;
```
myui:
```java
 private static double DELTA = 1E-5d;
```

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+
+    // total number of documents
+    // in the truly online setting, this can be an estimate of the maximum number of documents that could ever seen
+    private int D_ = 11102;
+
+    // defined by (tau0 + countEMStep)^(-kappa_)
+    private double rhot;
+
+    // positive value which downweights early iterations
+    private double tau0_ = 1020;
```
myui:
```java
@Nonnegative
private double tau0_ = 1020;
```

core/src/main/java/hivemall/lda/OnlineLDAModel.java
```diff
@@ -0,0 +1,464 @@
...
+                    lambdaBar_label[k] += docRatio * phi_.get(d).get(label)[k];
+                }
+                lambdaBar.put(label, lambdaBar_label);
+            }
+        }
+
+        // update lambda_
+        for (String label : lambda_.keySet()) {
+            float[] lambda_label = lambda_.get(label);
+            float[] lambdaBar_label;
```
myui:
```java
for(Map.Entry<String, float[]> e: lambda_.entrySet()) {
   String label = e.getKey();
   float[] lambda_label = e.getValue();
}
```

core/src/main/java/hivemall/lda/OnlineLDAModel.java
takuti: It's true, but plz wait until the final push after WIP condition. I will later check if the algorithm is implemented correctly, and a new variable e.g. `expElogTheta` might be introduced. If so, I guess `ElogTheta` is better in terms of both readability. It should be easy to understand a relationship between `ElogTheta` and its exponential.

takuti:
#### Requirements

```sql
drop temporary function if exists tokenize;
create temporary function tokenize as 'hivemall.tools.text.TokenizeUDF';

drop temporary function if exists is_stopword;
create temporary function is_stopword as 'hivemall.tools.text.StopwordUDF';

drop temporary function if exists feature;
create temporary function feature as 'hivemall.ftvec.FeatureUDF';

drop temporary function if exists lda;
create temporary function lda as 'hivemall.lda.OnlineLDAUDTF';
```

#### Sample query

```sql
with features as (
  select
    docid,
    feature(word, count(word)) as f
  from (
    select 1 as docid, "Fruits and vegetables are healthy." as doc
    union all
    select 2 as docid, "I like apples, oranges, and avocados. I do not like the flu or colds." as doc
  ) t1 LATERAL VIEW explode(tokenize(doc, true)) t2 as word
  where
    not is_stopword(word)
  group by
    docid, word
),
t as (
  select docid, collect_set(f) as words
  from features
  group by docid
)
select lda(words, "-topic 2 -iter 20") from t
;
```

#### Result

|topic | word   | score|
|:---:|:---:|:---:|
|0     | fruits | 0.33372128|
|0     | vegetables  |    0.33272517|
|0     | healthy | 0.33246377|
|0     | flu   |  2.3617347E-4|
|0     | apples | 2.1898883E-4|
|0     | oranges | 1.8161473E-4|
|0     | like   | 1.7666373E-4|
|0     | avocados  |      1.726186E-4|
|0     | colds  | 1.037139E-4|
|1     | colds  | 0.16622013|
|1     | avocados |       0.16618845|
|1     | oranges | 0.1661859|
|1     | like  |  0.16618414|
|1     | apples |  0.16616651|
|1     | flu   |  0.16615893|
|1     | healthy | 0.0012059759|
|1     | vegetables  |    0.0010818697|
|1     | fruits  | 6.080827E-4|

Clearly, topic0 corresponds to doc1, and topic1 represents doc2 topic words.

@myui Could you review whether this interface is sufficient?

From now, I will carefully check if the algorithm is implemented correctly in `OnlineLDAModel`.

myui:
@takuti What's the usage to predict topics?

```sql
WITH test as (
  select
    docid,
    word,
    count(word) as value
  from (
    select 1 as docid, "Fruits and vegetables are healthy." as doc
    union all
    select 2 as docid, "I like apples, oranges, and avocados. I do not like the flu or colds." as doc
  ) t1 LATERAL VIEW explode(tokenize(doc, true)) t2 as word
  where
    not is_stopword(word)
  group by
    docid, word
)
select
  docid,
  max_label(score, label) as label
from (
  select
    t.docid,
    m.label,
    sum(t.value * m.score) as score
  from
    test t
    JOIN lda_model m ON (t.word = m.word)
  group by
    t.docid, m.label
) t;
```

myui:
@takuti If you are going to create predict UDAF, better to rename to `lda` to `train_lda`.

Prediction expected to have feature vector as input and a predicted topic (& a posterior probabilities of topics) as outputs.

takuti:
@myui Thank you for the comments. Computing topic distribution for a document is a little bit complex, so I will implement UDAF for prediction by using `getTopicDistribution()` in 80a31539bf653e50471346777842ff9478ae352d

redsofa:
Thanks ! UDFS will make this much easier from what I understand. Another slightly related question... Is there a way to get at the intermediate topic lists for each mini-batch ?

takuti:
@myui I considered to design prediction UDAF, but IMO your suggestion above `sum(t.value * m.score) as score` is better for now. 

In order to compute topic distribution based on the `lambda` values (i.e., LDA model), I actually like to launch the E step for a test sample as [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/14031f6/sklearn/decomposition/online_lda.py#L546-L577) and current `getTopicDistribution()` (80a31539bf653e50471346777842ff9478ae352d) do. However, it requires prediction UDAF to know hyper-parameters (e.g., number of topics, alpha) which were used for training, and it's essentially infeasible. In addition, since users sometimes want to know posterior probabilities and their labels for each of all topics as follows, so single-column-output of UDAF is not sufficient.

| docid | label | prob |
|:---:|:---:|:---|
|1  |     0      | 0.9957867647115234
|1   |    1     |  0.004213235288476648
|2    |   0    |   0.0014898943734896843
|2     |  1   |    0.9985101056265103

See [HERE](https://gist.github.com/takuti/d24324e76d4b2ec7dc4b1d50a4d192d8) for detail.

Of course, since we do not launch the "expectation" step as theory suggests, `prob` is approximated value in some sense. But, I guess it's sufficient in practice.

takuti:
Or, aggregating over `docid` and output single row as an struct is possible:

| docid | probabilities |
|:---:|:---|
|1| { 0:0.9957867647115234, 1:0.004213235288476648 } |
|2| { 0:0.0014898943734896843, 1:0.9985101056265103 } |

Here, hyper-parameters used in the E step as follows:

- number of topics
- alpha (prior)
- delta (check convergence)

Fortunately, by counting unique `label` values, UDAF can know the number of topics when the `terminate` operation is launched. So, one possible direction is to create new `OnlineLDAModel` in UDAF's `terminate` w/ the number of topics, and we set default values to `alpha` and `delta` (or manually pass to UDAF as options).

myui:
```
WITH result as (
  select
    t.docid,
    lda_predict(t.wordid, t.value, m.label, m.lambda, '-alpha 0.01') 
      as probabilities -- list (struct<label,score>)
  from
    test t
    JOIN lda_model m ON (t.wordid = m.wordid)
  group by
    t.docid, m.label
)
select 
  docid,
  probabilities[0].label
from
  result;
```

myui:
TravisCI is failing.. 
```
test(hivemall.lda.OnlineLDAModelTest)  Time elapsed: 0.566 sec  <<< FAILURE!
java.lang.AssertionError: `vegetables` SHOULD be more suitable topic word than `flu` in topic 0
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at hivemall.lda.OnlineLDAModelTest.test(OnlineLDAModelTest.java:82)
```

Also, could you DEBUG print off in the unit tests?

takuti:
@myui okay, will add and check the failing later

takuti:
@myui Ready for review. [Usage](https://gist.github.com/takuti/d24324e76d4b2ec7dc4b1d50a4d192d8) has been updated.

takuti:
Since [castLazyBinaryObject()](https://github.com/takuti/incubator-hivemall/blob/810f5409eba8dff131e5d3b44069fb1182fa46cc/core/src/main/java/hivemall/utils/hadoop/HiveUtils.java#L932-L939) is helpful to improve readability, plz merge #63 before this PR if possible.

myui:
@takuti Could you change package from `hivemall.clustering` to `hivemall.topicmodel`?

takuti:
@myui Updated the package name, and, since #63 has been merged, applied some modifications. Again, this PR is ready for review 👍

myui:
@takuti Could you refine mini-batch implementation and perplexity computation scheme based on https://github.com/myui/incubator-hivemall/commit/cf84abc731fb00a33b7cec177850f8c026416cb0 ?

takuti:
^ cherry-picked except for unrelated modifications in AUC UDAF

core/src/main/java/hivemall/topicmodel/LDAUDTF.java
```diff
@@ -0,0 +1,553 @@
...
+            } catch (Throwable e) {
+                throw new UDFArgumentException(e);
+            }
+            this.inputBuf = buf = ByteBuffer.allocateDirect(1024 * 1024); // 1 MB
+            this.fileIO = dst = new NioStatefullSegment(file, false);
+        }
+
+        int wcLength = 0;
+        for (String wc : wordCounts) {
+            wcLength += wc.getBytes().length;
```
myui: NPE where wc is null

core/src/main/java/hivemall/topicmodel/LDAUDTF.java
```diff
@@ -0,0 +1,553 @@
...
+            model.setNumTotalDocs((int) count);
+        }
+
+        recordTrainSampleToTempFile(wordCounts);
+
+        miniBatch[miniBatchCount] = wordCounts;
+        miniBatchCount++;
+
+        if (miniBatchCount == miniBatchSize) {
+            model.train(miniBatch);
```
myui: null element handling for `miniBatch` is required

core/src/main/java/hivemall/topicmodel/LDAUDTF.java
```diff
@@ -0,0 +1,553 @@
...
+
+    @Override
+    public void process(Object[] args) throws HiveException {
+        int length = wordCountsOI.getListLength(args[0]);
+        String[] wordCounts = new String[length];
+        int j = 0;
+        for (int i = 0; i < length; i++) {
+            Object o = wordCountsOI.getListElement(args[0], i);
+            if (o == null) {
+                continue;
```
myui: throw an exception to avoid null elements (?)

myui:
revising at https://github.com/apache/incubator-hivemall/compare/master...myui:HIVEMALL-91-2

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -350,66 +348,80 @@ private float computeApproxBoundForMiniBatch() {
         float score = 0.f;
 
         // prepare
-        final float[] gammaSum = new float[_miniBatchSize];
-        for (int d = 0; d < gammaSum.length; d++) {
-            gammaSum[d] += MathUtils.sum(_gamma[d], _K);
+        float[] gammaSum = new float[_miniBatchSize];
+        Arrays.fill(gammaSum, 0.f);
```
myui: zero filling is not required in java. `gammaSum` has initialized with 0 for each entry.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -350,66 +348,80 @@ private float computeApproxBoundForMiniBatch() {
...
+        for (int d = 0; d < _miniBatchSize; d++) {
+            for (int k = 0; k < _K; k++) {
+                gammaSum[d] += _gamma[d][k];
```
myui: gammaSum[d] += MathUtils.sum(_gamma[d]);

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -350,66 +348,80 @@ private float computeApproxBoundForMiniBatch() {
...
+            for (String label : doc.keySet()) { // for each word in the document
+                float[] temp = new float[_K];
+                float max = Float.MIN_VALUE;
                 for (int k = 0; k < _K; k++) {
                     float eLogTheta_dk = (float) (Gamma.digamma(_gamma[d][k]) - Gamma.digamma(gammaSum[d]));
-                    float eLogBeta_kw = 0.f;
-                    if (_lambda.containsKey(d)) {
-                        eLogBeta_kw = (float) (Gamma.digamma(_lambda.get(d)[k]
-                                - Gamma.digamma(lambdaSum[k])));
+                    float eLogBeta_kw = (float) (Gamma.digamma(_lambda.get(label)[k])
```
myui: hash lookup `_lambda.get(label)` should be placed out of loop.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -350,66 +348,80 @@ private float computeApproxBoundForMiniBatch() {
...
+        // assuming likelihood for when corpus in the documents is only a subset of the whole corpus
+        // (i.e., online setting)
+        // likelihood is always roughly on the same scale
+        score *= _docRatio;
 
-                tmp += (-1.f * (float) Gamma.logGamma(lambdaSum[k]) - tmpPartial);
+        // E[log p(beta | eta) - log q (beta | lambda)]
+        float etaSum = _eta * _lambda.size(); // vocabSize * eta
+        for (int k = 0; k < _K; k++) {
+            for (String label : _lambda.keySet()) {
```
myui:
```java
 for (String label : _lambda.keySet()) {
    _lambda.get(label)
```

better to be replaced with

```java
for(float[] lambda: _lambda.values())
```

Avoid hash table lookups.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -350,66 +348,80 @@ private float computeApproxBoundForMiniBatch() {
...
+                gammaSum[d] += _gamma[d][k];
+            }
         }
-        final float[] lambdaSum = new float[_K];
-        for (float[] lambda_l : _lambda.values()) {
-            MathUtils.add(lambdaSum, lambda_l, _K);
+        float[] lambdaSum = new float[_K];
+        Arrays.fill(lambdaSum, 0.f);
+        for (int k = 0; k < _K; k++) {
+            for (String label : _lambda.keySet()) {
```
myui: can be revised to

```java
for(float[] lambda: _lambda.values()) {
   MathUtils.add(lambdaSum, lambda, _K);
```

takuti:
@myui Thank you for review. I'll add unit test which verifies perplexity computation, and double-check my implementation and your review later.

takuti:
Updating docs and testing on EMR...

takuti:
Manual tests both on local and EMR have been done. Could you review this? @myui

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,517 @@
...
+        }
+        final float[] lambdaSum = new float[_K];
+        for (float[] lambda : _lambda.values()) {
+            MathUtils.add(lambdaSum, lambda, _K);
+        }
+
+        for (int d = 0; d < _miniBatchSize; d++) {
+            // E[log p(doc | theta, beta)]
+            for (Map.Entry<String, Float> e : _miniBatchMap.get(d).entrySet()) {
+                final float[] lambda = _lambda.get(e.getKey());
```
myui: how about memoize `Gamma.digamma(lambda[k])` and `Gamma.digamma(lambdaSum[k]))` outside loop ` for (int d = 0; d < _miniBatchSize; d++) {`.

`lambda` and `lambdaSum` are stable unchanged in the loop. Also, they are only used for `digamma`.

```
MathUtils

@Nonnull
public static float[] digamma(@Nonnull final float[] lambda)
  final int k = lambda.length;
  final float[] ret = new float[k];
  for(int i=0; i<k; i++ {
     ret[i] = Gamma.digamma(lambda[k]);
  }
  return ret;
}
```

So, `digammaSum[d] = digamma(MathUtils.sum(_gamma[d]));`

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,517 @@
...
+        // assuming likelihood for when corpus in the documents is only a subset of the whole corpus
+        // (i.e., online setting); likelihood should be always roughly on the same scale
+        score *= _docRatio;
+
+        // E[log p(beta | eta) - log q (beta | lambda)]
+        float etaSum = _eta * _lambda.size(); // vocabulary size * eta
+        for (int k = 0; k < _K; k++) {
+            for (float[] lambda : _lambda.values()) {
+                // sum( (eta - lambda) * Elogbeta )
+                score += (_eta - lambda[k])
```
myui:
```java
final float lambdaK = lambda[k];
```
Array access is slow in the loop.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
takuti: For `Gamma.digamma(lambdaSum[k])` and `Gamma.digamma(gammaSum[d])`, I agree with your suggestion as follows:

```java
final float[] gammaSum = new float[_miniBatchSize];
for (int d = 0; d < _miniBatchSize; d++) {
    gammaSum[d] = MathUtils.sum(_gamma[d]);
}
final float[] digamma_gammaSum = MathUtils.digamma(gammaSum);

final float[] lambdaSum = new float[_K];
for (float[] lambda : _lambda.values()) {
    MathUtils.add(lambdaSum, lambda, _K);
}
final float[] digamma_lambdaSum = MathUtils.digamma(lambdaSum);
```

However, for `Gamma.digamma(lambda[k])`, I'm not sure if pre-computing is worthy of consideration. 

In the current implementation, the `Gamma.digamma` function is applied for `lambda` of each document-word pair:

```java
for (int d = 0; d < _miniBatchSize; d++) {
    // E[log p(doc | theta, beta)]
    for (Map.Entry<String, Float> e : _miniBatchMap.get(d).entrySet()) {
        final float[] lambda = _lambda.get(e.getKey());

        // logsumexp( Elogthetad + Elogbetad )
        for (int k = 0; k < _K; k++) {
           
            // call `Gamma.digamma(lambda[k])`
```

Thus, `Gamma.digamma(lambda[k])` is called `(mini-batch size) * (vocabulary size in each document) * k` times. 

By contrast, if we try to pre-compute `Gamma.digamma(lambda[k])`, it will be called `(whole vocabulary size) * k` times.

That is, if `(mini-batch size) * (vocabulary size in each document)` is greater than `(whole vocabulary size)`, your suggestion should be acceptable. How do you think about this? My concerns are:

- `mini-batch size` is generally small (1~100?)
- document-word matrices are generally sparse; each document only has a small fraction of `whole vocabulary size`
- => thus, `(mini-batch size) * (vocabulary size in each document)` could be smaller than `(whole vocabulary size)`

myui: @takuti hmm.. 

For `Gamma.digamma(lambda[k]`, it's okey as it is.

Please rename variable name `lambda` to `lambdaD`.

myui: I think

```java
final float[] gammaSum = new float[_miniBatchSize];
for (int d = 0; d < _miniBatchSize; d++) {
    gammaSum[d] = MathUtils.sum(_gamma[d]);
}
final float[] digamma_gammaSum = MathUtils.digamma(gammaSum);
```

can  be simplified as follows:

```java
final float[] digamma_gammaSum = new float[_miniBatchSize];
for (int d = 0, len = digamma_gammaSum.length; d < len; d++) {
    digamma_gammaSum[d] = Gamma.digamma(MathUtils.sum(_gamma[d]));
}
```

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,517 @@
...
+                        * (float) (Gamma.digamma(lambda[k]) - Gamma.digamma(lambdaSum[k]));
+
+                // sum( gammaln(lambda) - gammaln(eta) )
+                score += (float) (Gamma.logGamma(lambda[k]) - Gamma.logGamma(_eta));
```
myui: consider unchanged `logGamma` memoization for loops.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
takuti: @myui Sure, thanks for your opinion. `gammaSum[d]` is used at `score -= Gamma.logGamma(gammaSum[d]);`, so I will keep this.

core/src/main/java/hivemall/topicmodel/LDAUDTF.java
```diff
@@ -0,0 +1,567 @@
...
+                throw new HiveException("Given feature vector contains invalid elements");
+            }
+            String s = o.toString();
+            wordCounts[j] = s;
+            j++;
+        }
+
+        count++;
+        if (isAutoD) {
+            model.setNumTotalDocs((int) count);
```
myui: `model.setNumTotalDocs((int) count);`

int casting is not required.

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,530 @@
...
+public final class OnlineLDAModel {
+
+    // number of topics
+    private int _K;
+
+    // prior on weight vectors "theta ~ Dir(alpha_)"
+    private float _alpha = 1 / 2.f;
+
+    // prior on topics "beta"
+    private float _eta = 1 / 20.f;
```
myui: Better to fix values as final in constructor.

```java
    // number of topics
    private final int _K;

    // prior on weight vectors "theta ~ Dir(alpha_)"
    private final float _alpha;

    // prior on topics "beta"
    private final float _eta;
```

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,530 @@
...
+    private double _kappa = 0.7d;
+
+    // how many times EM steps are launched; later EM steps do not drastically forget old lambda
+    private long _updateCount = 0L;
+
+    // random number generator
+    @Nonnull
+    private final GammaDistribution _gd;
+    private static double SHAPE = 100.d;
+    private static double SCALE = 1.d / SHAPE;
```
myui: const value should be `static final`.

```java
private static final double SHAPE
private static final double SCALE
```

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,530 @@
...
+
+    // parameters
+    @Nonnull
+    private List<Map<String, float[]>> _phi;
+    private float[][] _gamma;
+    @Nonnull
+    private final Map<String, float[]> _lambda;
+
+    // check convergence in the expectation (E) step
+    private double _delta = 1e-5;
```
myui: should be fixed in constructor.

```java
private final double _delta;
```

core/src/main/java/hivemall/topicmodel/OnlineLDAModel.java
```diff
@@ -0,0 +1,530 @@
...
+    // defined by (tau0 + updateCount)^(-kappa_)
+    // controls how much old lambda is forgotten
+    private double _rhot;
+
+    // positive value which downweights early iterations
+    @Nonnegative
+    private double _tau0 = 1020;
+
+    // exponential decay rate (i.e., learning rate) which must be in (0.5, 1] to guarantee convergence
+    private double _kappa = 0.7d;
```
myui: `_tau0` and `_kappa` should be fixed as `final` in the constructor for readability.

Check other constructor parameters as well.

takuti:
**Note on the performance**

For [news20-multiclass](https://github.com/apache/incubator-hivemall/tree/master/core/src/test/resources/hivemall/classifier) data, I have translated [our Java test case](https://github.com/takuti/incubator-hivemall/blob/709848d5626f0df7e7361511224e0e9284b3484d/core/src/test/java/hivemall/topicmodel/OnlineLDAModelTest.java#L147-L223) to [Python scikit-learn implementation](https://github.com/takuti-sandbox/tmp/blob/57f740a3d0283e5586cc2cd170a8dd15b9cf96ac/python/lda/news20.py) w/ (almost) same setting.

In our Java code, unit test finishes in **8 sec** w/ approximately 30 iterations. By contrast, the Python implementation takes around **15 sec** for 30 iterations. Thus, even if `train_lda()` takes very long time for large-scale data, it should be natural. Hopefully, larger `-delta`, smaller `-iteration` or smaller `-eps` option could reduce running time (and end up w/ poor results).

* Python code actually creates and handles a 20-by-62061 huge, sparse matrix. It might be unfair, but Java code alternatively has many inefficient Map and Array accesses.

myui:
@takuti Merged w/ some refactoring. Great work! Thanks.

myui:
@NaokiStones Merged finally. FYI.

PR #68, opened 2017-04-07
[HIVEMALL-84] Add Docker Support

amaya382:
## What changes were proposed in this pull request?

Add Docker support for development and testing, including build tools for hivemall. Currently, only manual build is supported.

## What type of PR is it?

Feature, Documentation

## What is the Jira issue?

[HIVEMALL-84](https://issues.apache.org/jira/browse/HIVEMALL-84)

## How was this patch tested?

manual tests and automated build tests on DockerHub

## How to use this feature?

See [Document](https://raw.githubusercontent.com/apache/incubator-hivemall/9a3bc5c9aa5b6d26a4c93ba0140c367dc9431090/docs/gitbook/docker/getting_started.md)

~~Official automated build is future work but Dockerfile for automated build is ready. [Trial repository](https://hub.docker.com/r/amaya382/hivemall/)~~
**Known issue: To use automated build, we have to move Dockerfile into root directory**

PR #74, opened 2017-04-25
[HIVEMALL-84-2] Fix docker

amaya382:

Fix docker (https://github.com/apache/incubator-hivemall/pull/68#issuecomment-297029602)

## What type of PR is it?

Bug Fix

## What is the Jira issue?

[HIVEMALL-84](https://issues.apache.org/jira/browse/HIVEMALL-84)

## How was this patch tested?

manual tests

### DIFF
--- a/docs/gitbook/docker/getting_started.md
+++ b/docs/gitbook/docker/getting_started.md
@@ -35,9 +35,9 @@ This page introduces how to run Hivemall on Docker.
   
   `docker-compose -f resources/docker/docker-compose.yml build`
 
-## Build using docker engine
+## Build using docker command
   
-  `docker build -f resources/docker/Dockerfile`
+  `docker build -f resources/docker/Dockerfile .`
 
 # 2. Run container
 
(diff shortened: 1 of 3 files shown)


myui:
@amaya382 build of hivemall better to be optional in the document. Include ready-to-use jars on docker image.

I think Docker file should invoke `bin/build.sh` to create jars before `ln -s`.
https://github.com/apache/incubator-hivemall/blob/master/resources/docker/Dockerfile#L45

amaya382:
@myui I've wanted to write as option explicitly but it makes doc complex. because if use data volume (probably most frequent use case), building becomes a must. so in fact, it's necessary to build.
if distribute via dockerhub, we can divide use prebuild image (for users, don't need to build hivemall) and use from source (for developers, need to build hivemall), and can write easy document.
Wait for dockerhub support or not?

> I think Docker file should invoke bin/build.sh to create jars before ln -s
what will change?

myui:
@amaya382 

> what will change?

`${HIVEMALL_PATH}/target/hivemall-core-${HIVEMALL_VERSION}-with-dependencies.jar` is missing before building hivemall.

amaya382:
Now blocked by [HIVEMALL-100](https://issues.apache.org/jira/browse/HIVEMALL-100)

myui:
@amaya382 Merged with modidications. Thanks!

PR #85, opened 2017-06-08
[HIVEMALL-115] Fix ranking AUC for all TP/FP recommendation

takuti:

The current ranking AUC returns NULL when all recommended items are TP (or FP). This PR fixes the issue by returning 0.5 as "meaningless" AUC for such input.

- Caused by: myui/hivemall#326
- Reference:
  - [LibRec](https://github.com/guoguibing/librec/blob/18176ed41027348ee2187d8686a1b2c0d4d39277/core/src/main/java/net/librec/eval/ranking/AUCEvaluator.java#L80-L83)
  - [MyMediaLite](https://github.com/zenogantner/MyMediaLite/blob/3a268110338b24627c8c2c973c0f9d03b151e2df/src/MyMediaLite/Eval/Measures/AUC.cs#L51-L52)

## What type of PR is it?

Bug Fix

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-115

## How was this patch tested?

Manually (local & EMR)

### DIFF
--- a/core/src/main/java/hivemall/evaluation/BinaryResponsesMeasures.java
+++ b/core/src/main/java/hivemall/evaluation/BinaryResponsesMeasures.java
@@ -191,6 +191,11 @@ public final class BinaryResponsesMeasures {
         // # of all possible <TP, FP> pairs
         int nPairs = nTruePositive * (recommendSize - nTruePositive);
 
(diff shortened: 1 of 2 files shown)


myui:
LGTM 👍

PR #79, opened 2017-05-15
[HIVEMALL-101] Separate optimizer implementation

takuti:
## What changes were proposed in this pull request?

Finalize #14 

## What type of PR is it?

Improvement, Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-101

## How was this patch tested?

- Unit test
- Manual test on EMR

---

Todo:

- [x] Compare the accuracy of `-loss logloss` with the current `logress()` UDTF
- [x] Add document that includes general explanation of SGD, regularization and optimization techniques like [MLlib](http://spark.apache.org/docs/latest/mllib-optimization.html) and [sklearn](http://scikit-learn.org/stable/modules/sgd.html#mathematical-formulation)
- [x] Support `-mini_batch` option in a similar way to [what RegressionBaseUDTF does](https://github.com/takuti/incubator-hivemall/blob/5dc6f4eb5a8d8532201f6706673e2381d47d7e70/core/src/main/java/hivemall/regression/RegressionBaseUDTF.java#L247-L251); accumulate gradients over M samples, and update for mean value
- [x] ~~Save samples to external files as the other UDTFs (see LDA/pLSA UDTFs) and add `-iter` option~~ *This should be the other issue* [\[HIVEMALL-108\]](https://issues.apache.org/jira/browse/HIVEMALL-108)  
- [x] Manage cumulative loss for future `-iter` support
- [x] Anything else for mix-server-related things?

PR #71, opened 2017-04-13
[HIVEMALL-74] Implement pLSA

takuti:
## What changes were proposed in this pull request?

Implement (incremental) probabilistic latent semantic analysis (pLSA) algorithm:

- Original papers:
  - [Probabilistic Latent Semantic Indexing](http://dl.acm.org/citation.cfm?id=312649)
  - [Probabilistic Latent Semantic Analysis](http://www.iro.umontreal.ca/~nie/IFT6255/Hofmann-UAI99.pdf)
- Incremental variant which is implemented in this PR:
  - [Incremental Probabilistic Latent Semantic Analysis for Automatic Question Recommendation](https://pdfs.semanticscholar.org/b66e/c7faf2e4888503f7ad1537d284f350fb3e58.pdf)

## What type of PR is it?

Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-74

## How was this patch tested?

- unit tests
- manual tests on local

## How to use this feature?

Basically, interfaces are similar to Online LDA implemented in #66; we have `train_plsa()` and `plsa_predict()` functions.

For a `docs` table:

| docid | doc  |
|:---:|:---|
| 1  | "Fruits and vegetables are healthy." |
|2 | "I like apples, oranges, and avocados. I do not like the flu or colds." |

the following query successfully learns topics behind the two documents:

```sql
with word_counts as (
  select
    docid,
    feature(word, count(word)) as f
  from docs t1 lateral view explode(tokenize(doc, true)) t2 as word
  where
    not is_stopword(word)
  group by
    docid, word
),
plsa_model as (
  select
    train_plsa(feature, "-topic 2 -iter 10000 -eps 0.00001 -delta 0.00001 -alpha 0.00001") as (label, word, prob)
  from (
    select docid, collect_set(f) as feature
    from word_counts
    group by docid
  ) t
),
test as (
  select
    docid,
    word,
    count(word) as value
  from docs t1 LATERAL VIEW explode(tokenize(doc, true)) t2 as word
  where
    not is_stopword(word)
  group by
    docid, word
),
topic as (
  select
    t.docid,
    plsa_predict(t.word, t.value, m.label, m.prob, "-topic 2 -delta 0.01 -alpha 0.00001") as probabilities
  from
    test t
    JOIN plsa_model m ON (t.word = m.word)
  group by
    t.docid
)
select docid, probabilities, probabilities[0].label, m.words -- topic each document should be assigned
from topic t
join (
[...]

PR #87, opened 2017-06-20
[HIVEMALL-108] Support `-iter` option in generic predictors

takuti:

Support `-iter` option in both **generic classifier** and **regression** by buffering input samples, with an auxiliary `-tol` option.

Ref. #79 

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-108

## How was this patch tested?

- Unit test
  - Passed with the same test cases which were used for the previous "manual" iteration using for-loops
- Manual test on EMR
  - Same as what I did in #79

## How to use this feature?

Nothing has been changed except for the new options :)

---

### Notice

In order to check convergence, this PR adds `-tol` option; it is not `-eps` as pLSA #71 and LDA #66 have, because `-eps` already exists in generic predictors as an option for adagrad/adadelta. For the sake of unifying interface, `train_lda` and `train_plsa` might need to support `-tol` as alias of the existing `-eps` option. If we do so, it could be a part of [HIVEMALL-120](https://issues.apache.org/jira/browse/HIVEMALL-120).

### DIFF
--- a/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
+++ b/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
@@ -261,7 +261,7 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
             } catch (Throwable e) {
                 throw new UDFArgumentException(e);
             }
-            this.inputBuf = buf = ByteBuffer.allocateDirect(1024 * 1024 * 10); // 10 MB
+            this.inputBuf = buf = ByteBuffer.allocateDirect(1024 * 1024); // 1 MB
             this.fileIO = dst = new NioStatefullSegment(file, false);
         }
 
--- a/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
+++ b/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
@@ -177,8 +177,8 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
 
         this.iterations = Primitives.parseInt(cl.getOptionValue("iterations"), 10);
         if (iterations < 1) {
-            throw new UDFArgumentException(
-                "'-iterations' must be greater than or equals to 1: " + iterations);
+            throw new UDFArgumentException("'-iterations' must be greater than or equals to 1: "
+                    + iterations);
         }
 
         this.tol = Primitives.parseDouble(cl.getOptionValue("tolerance"), 1E-1d);
@@ -237,8 +237,8 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
         recordTrainSampleToTempFile(featureVector, target);
     }
 
-    protected void recordTrainSampleToTempFile(@Nonnull final FeatureValue[] featureVector, final float target)
-            throws HiveException {
+    protected void recordTrainSampleToTempFile(@Nonnull final FeatureValue[] featureVector,
+            final float target) throws HiveException {
         if (iterations == 1) {
             return;
         }
@@ -427,7 +427,7 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
 
         final Reporter reporter = getReporter();
         final Counters.Counter iterCounter = (reporter == null) ? null : reporter.getCounter(
-                "hivemall.GeneralLearnerBase$Counter", "iteration");
+            "hivemall.GeneralLearnerBase$Counter", "iteration");
 
         try {
             if (dst.getPosition() == 0L) {// run iterations w/o temporary file
--- a/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
+++ b/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
@@ -469,8 +469,8 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
                 }
                 buf.flip();
 
-                int iter = 2;
-                for (; iter <= iterations; iter++) {
+                for (int iter = 2; iter <= iterations; iter++) {
+                    cvState.next();
                     reportProgress(reporter);
                     setCounterValue(iterCounter, iter);
 
@@ -491,17 +491,17 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
                         batchUpdate();
                     }
 
-                    if (cvState.isConverged(iter, numTrainingExamples)) {
+                    if (cvState.isConverged(numTrainingExamples)) {
                         break;
                     }
                 }
                 logger.info("Performed "
-                        + Math.min(iter, iterations)
+                        + cvState.getCurrentIteration()
                         + " iterations of "
                         + NumberUtils.formatNumber(numTrainingExamples)
                         + " training examples on memory (thus "
-                        + NumberUtils.formatNumber(numTrainingExamples * Math.min(iter, iterations))
-                        + " training updates in total) ");
+                        + NumberUtils.formatNumber(numTrainingExamples
+                                * cvState.getCurrentIteration()) + " training updates in total) ");
             } else {// read training examples in the temporary file and invoke train for each example
                 // write training examples in buffer to a temporary file
                 if (buf.remaining() > 0) {
@@ -522,8 +522,8 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
                 }
 
                 // run iterations
-                int iter = 2;
-                for (; iter <= iterations; iter++) {
+                for (int iter = 2; iter <= iterations; iter++) {
+                    cvState.next();
                     setCounterValue(iterCounter, iter);
 
                     buf.clear();
@@ -577,17 +577,17 @@ public abstract class GeneralLearnerBaseUDTF extends LearnerBaseUDTF {
                         batchUpdate();
                     }
 
-                    if (cvState.isConverged(iter, numTrainingExamples)) {
+                    if (cvState.isConverged(numTrainingExamples)) {
                         break;
                     }
                 }
                 logger.info("Performed "
-                        + Math.min(iter, iterations)
+                        + cvState.getCurrentIteration()
                         + " iterations of "
                         + NumberUtils.formatNumber(numTrainingExamples)
                         + " training examples on a secondary storage (thus "
-                        + NumberUtils.formatNumber(numTrainingExamples * Math.min(iter, iterations))
-                        + " training updates in total)");
+                        + NumberUtils.formatNumber(numTrainingExamples
+                                * cvState.getCurrentIteration()) + " training updates in total)");
             }
         } catch (Throwable e) {
             throw new HiveException("Exception caused in the iterative training", e);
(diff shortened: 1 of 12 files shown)


myui:
Thanks! will review in this week.

myui:
I'm wondering to generalize `-iterations` and `-mini_batch` logic of GeneralLeanerBaseUDTF for BinaryOnlineClassifierUDTF and RegressionBaseUDTF as well.
https://gyazo.com/4ceaeb003ccad8587ca55e080ba6ccfa

It's not straightforward though.

myui:
I think `-tol` is hard to configure by average users.

It can be replaced with `ConversionState`.  It doubly checks that an iteration really converges.

takuti:
@myui Thank you for review and suggestion. I will work on them in a week.

myui:
There are several design options to consider:
 1) move `-iter` and `-mini-batch` functionalities to LeanerBaseUDTF
   => but, it cannot be used for MulticlassOnlineClassifier.. :-(
 2) make BinaryOnlineClassfierUDTF/RegressionBaseUDTF the subclass of GeneraLearnerBaseUDTF
   => regression bug may occur ..
 3) make Abstract IterativeLearnerBaseUDTF as the subclass of LearnerBaseUDTF
   => too complex object hierarchy (not good design)
 4) implement `-iter` and `-mini-batch` functionalities in BinaryOnlineClassfierUDTF/RegressionBaseUDTF
   => duplicate code, hard for maintenance
 5) Mixin iteration supoprt
   => callback is needed and not a pattern for mixin
 6) generalize `-iter` support for MulticlassOnelineClassifierUDTF
   => not straightforward design

Also, `-mini_batch` design should be well revised for online perceptron variants.

myui:
Applied some refactoring in https://github.com/myui/incubator-hivemall/commit/71b7c09fb2e603f91fb98753d9362d27128d66f5

It may exists bugs in the Optimizer implementation. Too large losses are found when doing early termination.

https://gyazo.com/c6f4155a465ecb3a7ea916a2ee68db85
https://gyazo.com/26b9c8d7ac1be42df8397df74c11d0e5

core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
```diff
@@ -318,66 +412,240 @@ protected void onlineUpdate(@Nonnull final FeatureValue[] features, final float
...
+                    }
+                    buf.rewind();
+
+                    if (is_mini_batch) { // Update model with accumulated delta
+                        batchUpdate();
+                    }
+
+                    logger.info("[iter " + iter + "] cumulative loss: " + cumLoss);
+
+                    if (Math.abs(cumLossPrev - cumLoss) < tol) {
```
myui: `cumLoss` is always 0 due to ` this.cumLoss = 0.d;`. So, this conversion check is not working properly.

takuti:
@myui Oh okay, will catch-up soon. 

Refactoring and modification on ongoing PR are personally confusing by the way...let us first discuss here, and use GitHub's "review" feature more...it's an OSS project!

myui:
It takes so long to reflect this kind of changes by review comments. Find commit logs for my points for review.

https://github.com/myui/incubator-hivemall/commit/ac062f2f5f8e0eb9ad093d4b92f478d791fcbbf4 is bug fix of ConversionState usage (cumulative loss is not reset at the beginning of 2nd iteration).

https://github.com/myui/incubator-hivemall/commit/b999196913874361b3e9e12c65418675a3070c23 avoided `-tol` and fixed to use existing ConversionState.

https://github.com/myui/incubator-hivemall/commit/94e16366d54b6fa4e58a7e78d9db16aea83eee3f revised variable scope in ConversionState.

https://github.com/myui/incubator-hivemall/commit/94e16366d54b6fa4e58a7e78d9db16aea83eee3f Fixed a bug in readFrom.

https://github.com/myui/incubator-hivemall/commit/65233ff69d653df432d88eb4ecbc9031a6de60c5 numSteps should be long to avoid overflow.

I guess `-tol` used in your PR was too small to find existing bugs in optimizer (not bugs in this PR).

takuti:
> It may exists bugs in the Optimizer implementation. Too large losses are found when doing early termination.

Too large losses are probably observed as a consequence of inappropriate hyper-parameters. In the last implementation, I tried to test all loss functions under the same `-eta0` and `-lambda` as:

```java
for (String opt : optimizers) {
    for (String reg : regularizations) {
        if (reg == "RDA" && opt != "AdaGrad") {
            continue;
        }

        for (String loss : lossFunctions) {
            String options = "-opt " + opt + " -reg " + reg + " -loss " + loss
                    + " -lambda 1e-6 -cv_rate 0.005 -iter 512";
```

However, since each loss function draws different curve, optimal `-eta0` and `-lambda` should be different.

Finding optimal hyper-parameters for each combination of `-opt` + `-reg` + `-loss` is really hard, so I've updated `assert` condition; a new version of `GeneralRegressionUDTFTest` simply checks if iterative training improves MAE to some degree, by comparing with MAE w/ initial state of model.

takuti:
> generalize -iterations and -mini_batch logic of GeneralLeanerBaseUDTF for BinaryOnlineClassifierUDTF and RegressionBaseUDTF as well

I considered a little, but, as you suggested many possible ways, this complex topic should be handled more carefully as a different issue. The first step will be to support `-iter` in each of `BinaryOnlineClassifierUDTF` and `RegressionBaseUDTF`.

If you are okay to treat them as a different issue, I am ready for review & merge this PR.

myui:
> Too large losses are probably observed as a consequence of inappropriate hyper-parameters. In the last implementation, I tried to test all loss functions under the same -eta0 and -lambda as:

@takuti Why test was succeeding at https://github.com/apache/incubator-hivemall/pull/87/commits/b2f0460f3c210c9afbb9cdf2c1247ec34e8c5db2 ?

takuti:
@myui That's why I repeatedly tried a wide variety of combinations and finally found `-lambda 1e-6 -tol 1e-3f -iter 512` leads successful. 

What `-tol` did is to compare the absolute difference between the last and current cumulative loss; that is, the initial implementation checked convergence based on **absolute difference**. Meanwhile, `-cv_rate` looks **relative difference**. Hence, in a precise sense, your commit feb9bb132afb8d81357f43cf2ee2d94f6e0aa9b6 is not "refactor," and I need to find a completely different optimal combination of hyper-parameters for testing.

myui:
> What -tol did is to compare the absolute difference between the last and current cumulative loss; that is, the initial implementation checked convergence based on absolute difference. Meanwhile, -cv_rate looks relative difference. Hence, in a precise sense, your commit feb9bb1 is not "refactor," and I need to find a completely different optimal combination of hyper-parameters for testing.

`absolute difference => relative difference` DOES NOT explain too large loss in the middle of iterations. Even without conversion checks, it was never converged. 

Iterations were converged when using absolute difference and `Assert.assertTrue(cumLoss / samplesList.size() < 0.5d);` was passing. Why?

myui:
Bug in https://github.com/apache/incubator-hivemall/pull/87/commits/0c0e4ea68187fa4b672a892774caaf74d5a5e058 caused https://github.com/apache/incubator-hivemall/commit/b2f0460f3c210c9afbb9cdf2c1247ec34e8c5db2 to pass the test (?).

myui:
> Bug in 0c0e4ea caused b2f0460 to pass the test (?).

@takuti This assumption was correct. 

`GeneralRegressionUDTFTest` is *illegally* passing when reverting fix in https://github.com/apache/incubator-hivemall/commit/0c0e4ea68187fa4b672a892774caaf74d5a5e058.

Fails on `GeneralClassifierUDTFTest` is due to relaxing conversion checks.

- `this.feature = new Text(NIOUtils.getString(src));` and `-cvRate 0.005`

Both `GeneralClassifierUDTFTest` and `GeneralRegressionUDTFTest` fails.

- `this.feature = new Text(NIOUtils.getString(src));` and `-cvRate 0.001`

Both `GeneralClassifierUDTFTest` and `GeneralRegressionUDTFTest` succeeds.

- `this.feature = NIOUtils.getString(src);` and `-cvRate 0.001`

`GeneralClassifierUDTFTest` succeeds but `GeneralRegressionUDTFTest` fails.

myui:
It might exist more complex potential bug in the iteration support.

In the first iteration, feature index can be `Text` but `String` in the following iterations.

takuti:
@myui good catch & nice work~ 

If I remember correctly, I previously noticed the difference between `Text` and `String` as label of `FeatureValue`, and thus I replaced `NIOUtils.getString(src)` with `new Text(NIOUtils.getString(src))`. Since the unit test has been passed at the moment, I misguidedly considered the problem has been fixed.

I will try to look into the problem more.

myui:
@takuti 

- store feature type in GeneralLearnerBaseUDTF
  https://github.com/takuti/incubator-hivemall/blob/95927f64104bf75d071c8665d64afb48837fcb02/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java#L222
- wrap feature writes to use type info
  https://github.com/takuti/incubator-hivemall/blob/95927f64104bf75d071c8665d64afb48837fcb02/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java#L305
- wrap feature reads to use type info
  https://github.com/takuti/incubator-hivemall/blob/95927f64104bf75d071c8665d64afb48837fcb02/core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java#L483
  
Refer https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/fm/FactorizationMachineUDTF.java#L665 and https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/fm/FactorizationMachineUDTF.java#L630 for example

takuti:
@myui Do you mean I need to replace `FeatureValue[]` with `Feature[]` and support a new option `-int_feature`, in a similar way to what `FactorizationMachineUDTF` does?

myui:
@takuti No. 

`enum FeatureType { JavaString, Text, JavaInt, WritableInt, WritableLong }`

• `this.featureType = resolveType();` in `initialize()`.
• wrap feature read/write using `featureType`.

takuti:
@myui If I understand correctly, the feature-type-related issue has been fixed 3c753bbdaaf2c145048fe885bb44cb62f7a08232

myui:
@takuti Thanks! will review.

core/src/main/java/hivemall/GeneralLearnerBaseUDTF.java
```diff
@@ -220,6 +235,23 @@ protected PrimitiveObjectInspector processFeaturesOI(@Nonnull ObjectInspector ar
         this.featureListOI = (ListObjectInspector) arg;
         ObjectInspector featureRawOI = featureListOI.getListElementObjectInspector();
         HiveUtils.validateFeatureOI(featureRawOI);
+        if (featureRawOI instanceof JavaStringObjectInspector) {
```
myui: We need to consider other possibilities for StringObjectInspector.
https://gyazo.com/4c7c3efb22f441db09f614457235b5c0

I'll handle this issue.

myui:
https://issues.apache.org/jira/browse/HIVEMALL-135 

Documentation about `-iter` option is lacking.

PR #89, opened 2017-06-23
[HIVEMALL-120] Refactor on LDA/pLSA's mini-batch & buffered iteration logic

takuti:

Refactor LDA/pLSA implementation for better mini-batch & buffered iteration logic

## What type of PR is it?

Refactoring

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-120

## How was this patch tested?

Unit test & manual test on EMR

## How to use this feature?

Nothing has been changed

### DIFF
--- a/core/src/main/java/hivemall/topicmodel/AbstractProbabilisticTopicModel.java
+++ b/core/src/main/java/hivemall/topicmodel/AbstractProbabilisticTopicModel.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package hivemall.topicmodel;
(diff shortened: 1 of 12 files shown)


takuti:
At first, I planned to replace `String[][] miniBatch` with things like `FeatureValue[][] miniBatch` similarly to the generic predictor implementation; that is, instead of dealing word counts as String e.g. "foo:10," I tried to represent them as `FeatureValue` for readability. However, since `PLSAPredictionUDAF` and `LDAPredictionUDAF` need to create OI for such word counts, Java standard string object is much more handy rather than `FeatureValue` objects.

Thus, document is still represented as an array (or list) of String for now.

takuti:
@myui How do you think about adding `-tol` option as alias of the `-eps` option as I mentioned #87 
?

Also, if you have further ideas which improve the topic modeling UDFs, let you share and we discuss them here!

myui:
@takuti Adding an alias `-tol` for `-eps` is a good idea.

myui:
No need to replace `String[][] miniBatch`. It already using `FeatureVector` for parsing feature vectors as thus it accept the input format of Hivemall's feature vector.

takuti:
> Adding an alias -tol for -eps is a good idea.

Sure.

>  it accept the input format of Hivemall's feature vector

Oh, it's not format matter by the way; `String` or `FeatureValue` in terms of type.

Current implementation repeatedly parses feature-value-formatted `String` at [HERE](https://github.com/takuti/incubator-hivemall/blob/bc4b9d3a13310d85c0fb4fc572d0338fefee0a8d/core/src/main/java/hivemall/topicmodel/AbstractProbabilisticTopicModel.java#L66-L69) for every time mini-batch is passed. If we directly pass a two-dimensional array of `FeatureValue` as word counts, we can avoid the unnecessary parse operations.

myui:
ah, parsing multiple times could be avoided.

myui:
Merged this PR. Create `[HIVEMALL-120-2]` for `-tol alias for -eps`.

takuti:
Oh, you move very fast :) 

> -tol alias

Sure~

takuti:
@myui Oh, you move very fast :) 

> -tol alias

Sure~

PR #50, opened 2017-02-24
[HIVEMALL-80] Fix incorrect typeInfo checks in evaluation UDAFs

takuti:
## What changes were proposed in this pull request?

Fix incorrect typeInfo checks in evaluation UDAFs. These were caused by [this PR](https://github.com/myui/hivemall/pull/326).

## What type of PR is it?

Bug Fix (?)

## What is the Jira issue?

[HIVEMALL-80](https://issues.apache.org/jira/browse/HIVEMALL-80)

## How was this patch tested?

N/A

## How to use this feature?

N/A

PR #94, opened 2017-07-04
[HIVEMALL-80] Allow array<struct> type as the first parameter to ndcg

uhyonc:

Fix for checking the parameters to ndcg to allow struct types (which allows graded response measures)

## What type of PR is it?

Bug Fix

## What is the Jira issue?

HIVEMALL-80

## How was this patch tested?

Tested on our Hive cluster.


takuti:
@uhyonc Good catch! (ref #50) I've also confirmed it works fine.

myui:
Better to write URL is the PR description to link with JIRA.
https://issues.apache.org/jira/browse/HIVEMALL-80

PR #97, opened 2017-07-11
[HIVEMALL-130] Support user-defined dictionary for `tokenize_ja`

takuti:

- Add a new argument to `tokenize_ja` which enables users to register user-defined dictionary
  - Value can be either `const array<string>` (for array of custom definitions) or `const string` (for URL pointing an external dictionary file)
  - Users need to follow [Kuromoji official user dictionary format](https://github.com/atilika/kuromoji/blob/909fd6b32bf4e9dc86b7599de5c9b50ca8f004a1/kuromoji-core/src/test/resources/userdict.txt) as `<word>,<result>,<read>,<class>`
- Update document and refactor accordingly

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-130

## How was this patch tested?

Manually tested both on local and EMR with the Kuromoji official sample file:

```
hive> select
    >   tokenize_ja("日本経済新聞", "normal"),
    >   tokenize_ja("日本経済新聞", "normal", array(), array(), "https://raw.githubusercontent.com/atilika/kuromoji/909fd6b32bf4e9dc86b7599de5c9b50ca8f004a1/kuromoji-core/src/test/res/userdict.txt"),
    >   tokenize_ja("日本経済新聞", "normal", array(), array(), array("日本経済新聞,日本 経済 新聞,ニホン ケイザイ シンブン,カスタム名詞", "関西国際空港,関西 国際 空港,カンサイ コクサイ クウコウ,テスト名詞"))
    > ;
OK
_c0     _c1     _c2
["日本経済新聞"]        ["日本","経済","新聞"]  ["日本","経済","新聞"]
Time taken: 2.094 seconds, Fetched: 1 row(s)
```

## How to use this feature?

As shown above.

## Checklist

### DIFF
--- a/core/src/main/java/hivemall/utils/io/BoundedInputStream.java
+++ b/core/src/main/java/hivemall/utils/io/BoundedInputStream.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package hivemall.utils.io;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Input stream which only supplies bytes up to a certain length
+ * Implementation is based on BoundedInputStream in Apache Commons IO
+ * https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/BoundedInputStream.html
+ */
+public final class BoundedInputStream extends InputStream {
+    private static final int EOF = -1;
+
+    private final InputStream is;
+    private final long max;
+    private long pos = 0L;
+    private long mark = EOF;
+
+    public BoundedInputStream(final InputStream is) {
+        this(is, EOF); // wrapped, but unlimited input stream
+    }
+
+    public BoundedInputStream(final InputStream is, final long size) {
+        this.is = is;
+        this.max = size;
+    }
+
+    @Override
+    public int read() throws IOException {
+        if (max >= 0 && pos >= max) {
+            return EOF;
+        }
+
+        int result = is.read();
+        this.pos += 1;
+
+        return result;
+    }
+
+    @Override
+    public int read(final byte[] b) throws IOException {
+        return read(b, 0, b.length);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        if (max >= 0 && pos >= max) {
+            return EOF;
+        }
+
+        long maxRead = max >= 0 ? Math.min(len, max - pos) : len;
+        int bytesRead = is.read(b, off, (int) maxRead);
+        if (bytesRead == EOF) {
+            return EOF;
+        }
+        this.pos += bytesRead;
+
+        return bytesRead;
+    }
+
+    @Override
+    public long skip(long n) throws IOException {
+        long toSkip = max >= 0 ? Math.min(n, max - pos) : n;
+        long skippedBytes = is.skip(toSkip);
+        this.pos += skippedBytes;
+        return skippedBytes;
+    }
+
+    @Override
+    public int available() throws IOException {
+        if (max >= 0 && pos >= max) {
+            return 0;
+        }
+        return is.available();
+    }
+
+    @Override
+    public void close() throws IOException {
+        is.close();
+    }
+
+    @Override
+    public synchronized void reset() throws IOException {
+        is.reset();
+        this.pos = mark;
+    }
+
+    @Override
+    public synchronized void mark(int readlimit) {
+        is.mark(readlimit);
+        this.mark = pos;
+    }
+
+    @Override
+    public boolean markSupported() {
+        return is.markSupported();
+    }
+
+}
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -40,7 +40,7 @@ public final class HttpUtils {
 
     @Nonnull
     public static InputStream getLimitedInputStream(@Nonnull HttpURLConnection conn,
-            @Nonnegative long size) throws IOException {
+            @Nonnegative long size) throws NullPointerException, IOException {
         InputStream is = conn.getInputStream();
         is = new LimitedInputStream(is, size);
         return is;
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -39,10 +39,10 @@ public final class HttpUtils {
     }
 
     @Nonnull
-    public static InputStream getBoundedInputStream(@Nonnull HttpURLConnection conn,
+    public static InputStream getLimitedInputStream(@Nonnull HttpURLConnection conn,
             @Nonnegative long size) throws IOException {
         InputStream is = conn.getInputStream();
-        is = new BoundedInputStream(is, size);
+        is = new LimitedInputStream(is, size);
         return is;
     }
 }
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package hivemall.utils.io;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+public final class HttpUtils {
+    private HttpUtils() {}
+
+    @Nonnull
+    public static HttpURLConnection getHttpURLConnection(@Nonnull String urlStr)
+            throws IllegalArgumentException, IOException {
+        if (!urlStr.startsWith("http://") && !urlStr.startsWith("https://")) {
+            throw new IllegalArgumentException("Unexpected url: " + urlStr);
+        }
+        URL url = new URL(urlStr);
+        return (HttpURLConnection) url.openConnection();
+    }
+}
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -18,8 +18,6 @@
  */
 package hivemall.utils.io;
 
-import org.apache.commons.io.input.BoundedInputStream;
-
 import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
 import java.io.IOException;
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -18,8 +18,12 @@
  */
 package hivemall.utils.io;
 
+import org.apache.commons.io.input.BoundedInputStream;
+
+import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 
@@ -34,5 +38,12 @@ public final class HttpUtils {
         }
         URL url = new URL(urlStr);
         return (HttpURLConnection) url.openConnection();
+    }
+
+    @Nonnull
+    public static InputStream getBoundedInputStream(@Nonnull HttpURLConnection conn, @Nonnegative long size) throws IOException {
+        InputStream is = conn.getInputStream();
+        is = new BoundedInputStream(is, size);
+        return is;
     }
 }
--- a/core/src/main/java/hivemall/utils/io/HttpUtils.java
+++ b/core/src/main/java/hivemall/utils/io/HttpUtils.java
@@ -41,7 +41,8 @@ public final class HttpUtils {
     }
 
     @Nonnull
-    public static InputStream getBoundedInputStream(@Nonnull HttpURLConnection conn, @Nonnegative long size) throws IOException {
+    public static InputStream getBoundedInputStream(@Nonnull HttpURLConnection conn,
+            @Nonnegative long size) throws IOException {
         InputStream is = conn.getInputStream();
         is = new BoundedInputStream(is, size);
         return is;
--- a/core/src/main/java/hivemall/utils/io/IOUtils.java
+++ b/core/src/main/java/hivemall/utils/io/IOUtils.java
@@ -33,6 +33,8 @@ import java.io.InputStreamReader;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
+import java.io.PushbackInputStream;
+import java.util.zip.GZIPInputStream;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -127,6 +129,29 @@ public final class IOUtils {
         final int ch3 = in.read();
         final int ch4 = in.read();
         return ((ch1 << 24) + (ch2 << 16) + (ch3 << 8) + (ch4 << 0));
+    }
+
+    /**
+     * Look ahead InputStream and decompress it as GZIPInputStream if needed
+     *
+     * @link https://stackoverflow.com/a/4818946
+     */
+    public static InputStream decompressStream(final InputStream is) throws IOException {
+        final PushbackInputStream pb = new PushbackInputStream(is, 2);
+
+        // look ahead
+        int b1 = pb.read();
+        int b2 = pb.read();
+
+        // push back
+        pb.unread(b2);
+        pb.unread(b1);
+
+        if (((b2 << 8) | b1) == GZIPInputStream.GZIP_MAGIC) {
+            return new GZIPInputStream(pb);
+        } else {
+            return pb;
+        }
     }
 
     public static void writeChar(final char v, final OutputStream out) throws IOException {
--- a/core/src/main/java/hivemall/utils/io/LimitedInputStream.java
+++ b/core/src/main/java/hivemall/utils/io/LimitedInputStream.java
@@ -35,9 +35,9 @@ public class LimitedInputStream extends FilterInputStream {
     protected final long max;
     protected long pos = 0L;
 
-    public LimitedInputStream(final InputStream is, @Nonnegative final long size) {
-        super(is);
-        this.max = size;
+    public LimitedInputStream(final InputStream in, @Nonnegative final long maxSize) {
+        super(in);
+        this.max = maxSize;
     }
 
     protected void raiseError() throws IOException {
@@ -53,29 +53,29 @@ public class LimitedInputStream extends FilterInputStream {
 
     @Override
     public int read() throws IOException {
-        int result = super.read();
-        if (result != -1) {
+        int res = super.read();
+        if (res != -1) {
             proceed(1L);
         }
-        return result;
+        return res;
     }
 
     @Override
     public int read(byte[] b, int off, int len) throws IOException {
-        int bytesRead = super.read(b, off, len);
-        if (bytesRead > 0) {
-            proceed(bytesRead);
+        int res = super.read(b, off, len);
+        if (res > 0) {
+            proceed(res);
         }
-        return bytesRead;
+        return res;
     }
 
     @Override
     public long skip(long n) throws IOException {
-        long skippedBytes = super.skip(n);
-        if (skippedBytes > 0) {
-            proceed(skippedBytes);
+        long res = super.skip(n);
+        if (res > 0) {
+            proceed(res);
         }
-        return skippedBytes;
+        return res;
     }
 
     @Override
--- a/core/src/main/java/hivemall/utils/io/LimitedInputStream.java
+++ b/core/src/main/java/hivemall/utils/io/LimitedInputStream.java
@@ -18,6 +18,9 @@
  */
 package hivemall.utils.io;
 
+import hivemall.utils.lang.Preconditions;
+
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnegative;
 import java.io.FilterInputStream;
 import java.io.IOException;
@@ -34,8 +37,9 @@ public class LimitedInputStream extends FilterInputStream {
     protected final long max;
     protected long pos = 0L;
 
-    public LimitedInputStream(final InputStream in, @Nonnegative final long maxSize) {
+    public LimitedInputStream(@CheckForNull final InputStream in, @Nonnegative final long maxSize) {
         super(in);
+        Preconditions.checkNotNull(in, "Base input stream must not be null");
         this.max = maxSize;
     }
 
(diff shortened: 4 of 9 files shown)


takuti:
^ In 474027264121c66a673cd6da483111b22f65b2fd, `tokenize_ja()` supports "NULL" stop words/tags in order to allow users to explicitly choose default set. That is, `null` and `array()` are dealt differently. This modification improves usability of the fifth argument.

```sql
hive> select
    >   tokenize_ja("今日は特にいい天気ですね。", "normal") as default1,
    >   tokenize_ja("今日は特にいい天気ですね。", "normal", null) as default2, -- use default stop words
    >   tokenize_ja("今日は特にいい天気ですね。", "normal", array()) as no_stopwords, -- "empty" (i.e., no) stop words
    >   tokenize_ja("今日は特にいい天気ですね。", "normal", null, null) as default3, -- use default stop word and tags
    >   tokenize_ja("今日は特にいい天気ですね。", "normal", null, array()) as no_stoptags, -- "empty" stop tags
    >   tokenize_ja("今日は特にいい天気ですね。", "normal", array(), array()) as no_stopwords_stoptags -- "empty" stop words and tags
    > ;
OK
default1        default2        no_stopwords    default3        no_stoptags     no_stopwords_stoptags
["今日","いい","天気"]  ["今日","いい","天気"]  ["今日","特に","いい","天気"]   ["今日","いい","天気"]  ["今日","いい","天気","ね"]     ["今日","は","特に","いい","天気","です","ね"]
```

takuti:
Note: I'm planning to make further improvement on the document before merging this PR

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -239,6 +239,16 @@ private static UserDictionary userDictionary(@Nullable final String userDictURL)
...
         conn.setReadTimeout(READ_TIMEOUT_MS); // throw exception from getXXX() methods
 
+        final int responseCode;
+        try {
+            responseCode = conn.getResponseCode();
+        } catch (Throwable e) {
+            throw new UDFArgumentException("Failed to get response code: " + e);
+        }
+        if (responseCode != 200) {
+            throw new UDFArgumentException("God invalid response code: " + responseCode);
```
takuti: OMG, it's typo lol

nlp/src/test/java/hivemall/nlp/tokenizer/KuromojiUDFTest.java
```diff
@@ -192,6 +241,126 @@ public void prepare(int arg) throws HiveException {}
...
+        // stopWords
+        argOIs[2] = ObjectInspectorFactory.getStandardConstantListObjectInspector(
+            PrimitiveObjectInspectorFactory.writableStringObjectInspector, null);
+        // stopTags
+        argOIs[3] = ObjectInspectorFactory.getStandardConstantListObjectInspector(
+            PrimitiveObjectInspectorFactory.writableStringObjectInspector, null);
+        // userDictArray (from https://raw.githubusercontent.com/atilika/kuromoji/909fd6b32bf4e9dc86b7599de5c9b50ca8f004a1/kuromoji-core/src/test/resources/userdict.txt)
+        List<String> userDict = new ArrayList<String>();
+        userDict.add("日本経済新聞,日本 経済 新聞,ニホン ケイザイ シンブン,カスタム名詞");
+        argOIs[4] = argOIs[2] = ObjectInspectorFactory.getStandardConstantListObjectInspector(
```
takuti: Wrong assignment

myui:
@takuti please change PR tile `[WIP]`

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -44,47 +53,52 @@
...
 @Description(
         name = "tokenize_ja",
-        value = "_FUNC_(String line [, const string mode = \"normal\", const list<string> stopWords, const list<string> stopTags])"
+        value = "_FUNC_(String line [, const string mode = \"normal\", const array<string> stopWords, const array<string> stopTags, const array<string> userDict (or string userDictURL)])"
                 + " - returns tokenized strings in array<string>")
 @UDFType(deterministic = true, stateful = false)
 public final class KuromojiUDF extends GenericUDF {
+    private static final int CONNECT_TIMEOUT_MS = 10000;
+    private static final int READ_TIMEOUT_MS = 30000;
+    private static final long MAX_INPUT_STREAM_SIZE = 1000000L; // ~1MB
```
myui: 1MB would be too small for a dict.

How about 32 * 1024 * 1024？

Read timeout then could be 60 sec or so considering a slow network.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,100 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+    private static UserDictionary userDictionary(@Nullable final String[] userDictArray)
+            throws UDFArgumentException {
+        if (userDictArray == null) {
+            return null;
+        }
+
+        StringBuilder builder = new StringBuilder();
+        for (String row : userDictArray) {
+            builder.append(row);
+            builder.append("\r\n");
```
myui: Is `\r ` reqeuired?

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,100 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+        }
+        InputStream is = new ByteArrayInputStream(builder.toString().getBytes());
```
myui: please use `FastByteArrayInputStream` .

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,100 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+        } catch (ZipException e) { // HTTP connection indicates a non-GZIP binary file
+            throw new UDFArgumentException(
+                "User dictionary URL MUST indicate plain text or GZIP file: " + e);
+        } catch (Throwable e) {
+            throw new UDFArgumentException("Failed to get input stream from the connection: " + e);
+        }
+
+        final Reader reader = new InputStreamReader(is);
+        try {
+            return UserDictionary.open(reader); // return null if empty
```
myui: should throw exception for null.

core/src/main/java/hivemall/utils/io/HttpUtils.java
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package hivemall.utils.io;
+
+import org.apache.commons.io.input.BoundedInputStream;
```
myui: Should not depend on commons-io for minimum dependencies.

Create it on `utils.io` .

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,100 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+            conn = HttpUtils.getHttpURLConnection(userDictURL);
+        } catch (Throwable e) {
+            throw new UDFArgumentException("Failed to create HTTP connection to the URL: " + e);
+        }
+
+        // allow to read as a compressed GZIP file for efficiency
+        conn.setRequestProperty("Accept-Encoding", "gzip");
+
+        conn.setConnectTimeout(CONNECT_TIMEOUT_MS); // throw exception from connect()
+        conn.setReadTimeout(READ_TIMEOUT_MS); // throw exception from getXXX() methods
```
myui: typo in the comment.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,100 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+        if (!textContent && !binaryContent) {
+            throw new UDFArgumentException(
+                "User dictionary URL indicates unexpected content type: " + contentType);
+        }
+
+        InputStream is;
+        try {
+            is = HttpUtils.getBoundedInputStream(conn, MAX_INPUT_STREAM_SIZE);
+            if ((textContent && "gzip".equals(conn.getContentEncoding())) || binaryContent) {
+                is = new GZIPInputStream(is);
```
myui: application/octet-stream != gzip file

takuti: I was not able to find a good way to explicitly detect "gzip" file (checking a magic number is one possible way though). If `InputStream` of binary file (i.e., `octet-stream`) is non-gzip one, `new GZIPInputStream` automatically throws an exception, and the later `catch (ZipException e)` handles the exception. 

https://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: Sure

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: Not necessarily. It's just CRLF.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: NULL user dictionary is allowed; it's a case that reader is empty.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: where?

myui: Was reviewing a stale commit. already fixed.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
myui: simple as follows:

```java
if(urlString.endWith(".gz")) {
   InputStream is = conn.getInputStream();
   is = new LimitedInputStream(is, MAX_SIZE);
   is = new GzipInputStream(is);
} else {
   conn.setRequestProperty("Accept-Encoding", "gzip");
   // conn.connect();
   InputStream is = conn.getInputStream();
   is = new LimitedInputStream(is, MAX_SIZE);
   if ("gzip".equals(con.getContentEncoding())) {     
     is = new GzipInputStream(is);
   }
}
```

takuti: Since URL can be easily impersonated, I don't accept to use url string for the purpose of detecting file type. However, for double-check purpose, your suggestion `if(urlString.endWith(".gz"))` is partially useful.

myui: MIME type is resolved by file appendix.
https://help.github.com/articles/mime-types-on-github-pages/

And, the above code works for shortened URLs.

Dictionary format should support ".txt", ".dict", and etc. I'm not sure `plain/text` is enough.
Special handling is only required for `.gz` and throw exception for malformed dictionary is enough.

Also, http servers do not send proper content type. For example, S3 need to set Content-type explicitly. So, depending on unreliable Content-type should be avoided.
http://blog.viktorkelemen.com/2012/05/amazon-s3-and-resource-interpreted-as.html

myui: https://www.google.com/search?q=Resource+interpreted+as+Stylesheet+but+transferred+with+MIME+type

core/src/main/java/hivemall/utils/io/BoundedInputStream.java
```diff
@@ -0,0 +1,117 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Input stream which only supplies bytes up to a certain length
+ * Implementation is based on BoundedInputStream in Apache Commons IO
+ * https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/BoundedInputStream.html
+ */
+public final class BoundedInputStream extends InputStream {
```
myui: `BoundedInputStream` returns at max N bytes. It's not what we want.

Should throw an exception when reached max bytes as seen in this [LimitedInputStream](https://github.com/apache/commons-fileupload/blob/master/src/main/java/org/apache/commons/fileupload/util/LimitedInputStream.java).

[Fail-fast](https://en.wikipedia.org/wiki/Fail-fast) behavior is expected instead of malformed stream cause exception on GzipInpuStream or Dictionary reader.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -215,7 +215,7 @@ private static UserDictionary userDictionary(@Nullable final String[] userDictAr
         StringBuilder builder = new StringBuilder();
         for (String row : userDictArray) {
             builder.append(row);
-            builder.append("\r\n");
+            builder.append(System.getProperty("line.separator"));
```
myui: I guess `\n` is enough for IPA dict format. 

In some OS, it can be `\r` and it thus be malformed.

takuti: Sure~

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: Thank you for the comments. Ah-hah, S3! It's true. I'm re-thinking here.

> And, the above code works for shortened URLs.

What do u mean? If URL is shortened, file extension is disappeared, and `urlString.endWith` does not work.

> Dictionary format should support ".txt", ".dict", and etc. I'm not sure plain/text is enough.

For non-gzipped files, `UserDictionary.open(reader)` throws an exception if the file is not in CSV format as a Kuromoji dictionary. So, we can skip checking `text/plain` and their file extension as you suggested in the above snippet; everything `.txt`, `.dic`, `.csv` is acceptable as long as its content is readable as a CSV file.

> Special handling is only required for .gz and throw exception for malformed dictionary is enough.

Yep. `new GZIPInputStream()` internally checks header magic number. We can utilize it somehow. I'm working on here under your suggestion.

core/src/main/java/hivemall/utils/io/BoundedInputStream.java
takuti: Ack! Good point. I'll re-implement it so.

> Fail-fast behavior is expected instead of malformed stream, causing an exception on GzipInputStream or Dictionary reader.

Could you explain a bit more detail?

myui: checksum of gzip is correct but your Boundedinputstream Could return unexpected eof.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
myui: For a shortened url, a gzip file could be gzipped again but it works and the rare case can be avoided.

myui: For certain cases, mime type of a csv file becomes non plain/text.

core/src/main/java/hivemall/utils/io/BoundedInputStream.java
takuti: Sry, till unclear. Could you check the latest code and demonstrate better implementation if possible?

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
takuti: Could you check the latest version and point out if something is still wrong or uncomfortable for you?

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
```diff
@@ -0,0 +1,106 @@
...
+    private void proceed(@Nonnegative long bytes) throws IOException {
+        this.pos += bytes;
+        if (pos > max) {
+            throw new IOException("Exceeded maximum size of input stream [" + max + " bytes]");
+        }
+    }
+
+    @Override
+    public int read() throws IOException {
+        int result = is.read();
```
myui: check [InputStream.read()](http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read())

no need to `proceed(1L)` as seen in

https://github.com/apache/commons-fileupload/blob/master/src/main/java/org/apache/commons/fileupload/util/LimitedInputStream.java#L100

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.InputStream;
+
+/**
+ * Input stream which is limited to a certain length
+ * Implementation is based on BoundedInputStream in Apache Commons IO and LimitedInputStream in Apache Commons FileUpload
+ *
+ * @link https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/BoundedInputStream.html
+ * @link https://commons.apache.org/proper/commons-fileupload/apidocs/org/apache/commons/fileupload/util/LimitedInputStream.html
+ */
+public final class LimitedInputStream extends InputStream {
```
myui: I like the original design of https://github.com/apache/commons-fileupload/blob/master/src/main/java/org/apache/commons/fileupload/util/LimitedInputStream.java#L65 that `raiseError` can override (customizable).

So, make `LimitedInputStream` class not final and create the default `raiseError` method.

```java
    protected void raiseError(long pos) throws IOException {
       throw new IOException("Exceeded maximum size of input stream: max=" + max ", pos="+ pos);
    }

    private void proceed(@Nonnegative long bytes) throws IOException {
         this.pos += bytes;
         if (pos > max) {
             raiseError(pos);
         }
    }
```

Example usage is as follows:

```java
new LimitedInputStream() {
   @Override
   protected void raiseError(long pos) throws IOException {
      logger.error(xxx);
      throw new IOException(...);
   }
};
```

myui: Why you LimitedInputStream is based on BoundedInputStream that extends InputStream?

[LimitedInputStream](
https://github.com/apache/commons-fileupload/blob/master/src/main/java/org/apache/commons/fileupload/util/LimitedInputStream.java#L27) that extends [FilterInputStream](https://docs.oracle.com/javase/jp/6/api/java/io/FilterInputStream.html) seems a better design for wrapping InputStream.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -170,6 +192,86 @@ private static CharArraySet stopWords(@Nonnull final String[] array)
...
+    private static UserDictionary userDictionary(@Nullable final String userDictURL)
+            throws UDFArgumentException {
+        if (userDictURL == null) {
+            return null;
+        }
+
+        final HttpURLConnection conn;
+        try {
+            conn = HttpUtils.getHttpURLConnection(userDictURL);
+        } catch (Throwable e) {
```
takuti: `Exception` is enough to catch both `IOException` and `IllegalArgumentException`

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
takuti: Make sense. Updating so.

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
takuti: Really? Proceeding 1 byte corresponds to `count++;` at L101 of the code.

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
```diff
@@ -0,0 +1,89 @@
...
+        long res = super.skip(n);
+        if (res > 0) {
+            proceed(res);
+        }
+        return res;
+    }
+
+    @Override
+    public int available() throws IOException {
+        if (pos >= max) {
```
myui: `available()` itself is not need to override.
`raiseError()` should be called when the max threshold is exceeded.

Some clients tried to sleep while input stream is filled as follows:

```java
while(true) {
   if(is.available() == 0) {// could return 0 where pos = max      
      if(numtry++ >= 2)  throw new IllegalStateException(xxxx);
      Time.sleep(1000); // sleep until filled
      continue;
   }
   // never reached here in the current code
   int bytes = in.read(buf); // expected to call raiseError where pos > max
   ...   
}
```

core/src/main/java/hivemall/utils/io/LimitedInputStream.java
```diff
@@ -0,0 +1,89 @@
...
+package hivemall.utils.io;
+
+import javax.annotation.Nonnegative;
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Input stream which is limited to a certain length
+ * Implementation is based on BoundedInputStream in Apache Commons IO and LimitedInputStream in Apache Commons FileUpload
```
myui: No need to put a link for BoundedInputStream. Comment can be fixed as follows:

```
Input stream which limits the bytes to read.
Implementation is based on LimitedInputStream of Apache Commons FileUpload.
````

File: core/src/main/java/hivemall/utils/io/LimitedInputStream.java (hunk 15 above)
Comment (takuti) [inline core/src/main/java/hivemall/utils/io/LimitedInputStream.java]: good point 👍

File: core/src/main/java/hivemall/utils/io/LimitedInputStream.java [hunk 17]
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @link https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/BoundedInputStream.html
+ * @link https://commons.apache.org/proper/commons-fileupload/apidocs/org/apache/commons/fileupload/util/LimitedInputStream.html
+ */
+public class LimitedInputStream extends FilterInputStream {
+
+    protected final long max;
+    protected long pos = 0L;
+
+    public LimitedInputStream(final InputStream in, @Nonnegative final long maxSize) {
```
myui: `Precondition.checkNotNull(in, "Base input stream must not be null");`

was better.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
myui: So then ` } catch (IOException | IllegalArgumentException  e) {`

http://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html

takuti: Thanks!

core/src/main/java/hivemall/utils/io/IOUtils.java
```diff
@@ -129,6 +131,29 @@ public static int readInt(final InputStream in) throws IOException {
...
+    /**
+     * Look ahead InputStream and decompress it as GZIPInputStream if needed
+     *
+     * @link https://stackoverflow.com/a/4818946
+     */
+    public static InputStream decompressStream(final InputStream is) throws IOException {
+        final PushbackInputStream pb = new PushbackInputStream(is, 2);
+
+        // look ahead
+        int b1 = pb.read();
```
myui: a potential bug.  `read()` could return -1 on EOF.

Here is a revised version.

```java
        // look ahead
        final byte[] signature = new byte[2];
        // If no byte is available because the stream is at the end of the file, the value -1 is returned; 
        // otherwise, at least one byte is read and stored into b.
        final int nread = pb.read(signature);
        if (nread > 0) {// may be -1 (EOF) or 1 or 2
            pb.unread(signature, 0, nread); // push back 
        }

        final int streamHeader = ((int) signature[0] & 0xff) | ((signature[1] << 8) & 0xff00);
        if (streamHeader == GZIPInputStream.GZIP_MAGIC) {
            return new GZIPInputStream(pb);
        } else {
            return pb;
        }
```

core/src/main/java/hivemall/utils/io/IOUtils.java
```diff
@@ -129,6 +131,29 @@ public static int readInt(final InputStream in) throws IOException {
...
+        int b2 = pb.read();
+
+        // push back
+        pb.unread(b2);
+        pb.unread(b1);
+
+        if (((b2 << 8) | b1) == GZIPInputStream.GZIP_MAGIC) {
```
myui: a potential bug. 

should be..
```java
final int streamHeader = ((int) b1 & 0xff) | ((b2 << 8) & 0xff00);
if (streamHeader == GZIPInputStream.GZIP_MAGIC) {
```

myui:
Merged with modifications. Thanks!

PR #103, opened 2017-07-20
[HIVEMALL-133][SPARK][WIP] Support spark-v2.2 in the hivemalls-spark module

maropu:
This pr added supports for spark-2.2 in Hivemall.

This pr is currently WIP because:
1. Java7's been dropped in spark-v2.2 and Hivemall still supports it, so we need some entries to check a Java version when `spark-2.2` enabled.
2. We need to move common code into `spark/spark-common`.

## What type of PR is it?
Improvement

## What is the Jira issue?
https://issues.apache.org/jira/browse/HIVEMALL-133

## How was this patch tested?
Existing tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,10 +34,7 @@ notifications:
   email: false
 
 script:
-  - mvn -q scalastyle:check test -Pspark-2.1
-  # test the spark-2.0 modules only in the following runs
-  - if [[ ! -z "$(java -version 2>&1 | grep 1.8)" ]]; then mvn -q scalastyle:check clean -Pspark-2.2 -Djava.source.version=1.8 -Djava.target.version=1.8 -pl spark/spark-2.2 -am test -Dtest=none; fi
-  - mvn -q scalastyle:check clean -Pspark-2.0 -pl spark/spark-2.0 -am test -Dtest=none
+  - ./bin/run_travis_tests.sh
 
 after_success:
   - mvn clean cobertura:cobertura coveralls:report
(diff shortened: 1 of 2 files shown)


myui:
We cannot go Java 8 yet since the stable version of Hadoop still depends on Java 7.

Better to keep minimum dependencies to Java 7 for Hivemall core module.

Also, `-target 1.7` is not supported in JDK 8.
https://stackoverflow.com/questions/21014338/can-you-compile-a-class-using-java-8-and-then-use-it-in-java-7

maropu:
Probably, we need to set both `-target 1.8` and `-Pspark-2.2` (also, we need to enforce users to do so) when we compile them for Spark-v2.2. If users set `-Pspark-2.2` only, it prints errors, or something.

myui:
I guess we can use different `javac` for each Maven module.
https://maven.apache.org/plugins/maven-compiler-plugin/examples/compile-using-different-jdk.html

myui:
@maropu see https://github.com/apache/incubator-hivemall/pull/105 for travis CI error.

Would be fixed when https://github.com/apache/incubator-hivemall/pull/105 is merged into master and rebase to master.

maropu:
ok

myui:
@maropu I would like to merge this PR but it seems that you deleted the branch or repository.

maropu:
Thanks, I'll check later

myui:
@maropu check #106 as well.

maropu:
See #112

PR #110, opened 2017-08-28
[HIVEMALL-142] Implement SingularizeUDF

takuti:

Implement `singularize(string word)` to obtain singular form of `word`.

The implementation referred the following third-party code:

- https://github.com/sundrio/sundrio/blob/95c2b11f7b842bdaa04f61e8e338aea60fb38f70/codegen/src/main/java/io/sundr/codegen/functions/Singularize.java
- https://github.com/clips/pattern/blob/3eef00481a4555331cf9a099308910d977f6fc22/pattern/text/en/inflect.py#L445-L623

## What type of PR is it?

Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-142

## How was this patch tested?

unit test & manual test on EMR

## How to use this feature?

as documented

## Checklist

### DIFF
--- a/core/src/main/java/hivemall/utils/lang/StringUtils.java
+++ b/core/src/main/java/hivemall/utils/lang/StringUtils.java
@@ -185,16 +185,14 @@ public final class StringUtils {
     public static String join(@Nonnull final List<String> list, @Nonnull final String sep) {
         final StringBuilder buf = new StringBuilder(128);
         for (int i = 0, size = list.size(); i < size; i++) {
-            final String s = list.get(i);
-            if (s == null) {
-                continue;
-            }
-
             if (i > 0) { // append separator before each element, except for the head element
                 buf.append(sep);
             }
 
-            buf.append(s);
+            final String s = list.get(i);
+            if (s != null) {
+                buf.append(s);
+            }
         }
         return buf.toString();
     }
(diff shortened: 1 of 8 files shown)


core/src/main/java/hivemall/utils/lang/StringUtils.java
```diff
@@ -172,12 +172,17 @@ public static void clear(@Nonnull final StringBuilder buf) {
 
     public static String concat(@Nonnull final List<String> list, @Nonnull final String sep) {
```
takuti: @myui I guess you originally assumed this method behaves in a similar way to [what `org.apache.commons.lang3.StringUtils.join` does](https://github.com/apache/commons-lang/blob/1571050a196198f336ae487ee3b6df629d3ee9da/src/main/java/org/apache/commons/lang3/StringUtils.java#L4106-L4150). However, the original code appends a separator even at the end of result string as:

- expected: `concat(["a", "b", "c"], "-")` => `a-b-c`
- actual: `concat(["a", "b", "c"], "-")` => `a-b-c-`

So, I fixed the method in 796d388c36c520858b6e61deb34100cb9201e5fa. Is this okay? If my assumption was incorrect, I revert the modification and introduce alternative method `StringUtils.join()`.

myui: former is expected.

takuti: okay, so I keep the update here

takuti: done

myui:
👍  Merged. Thanks!

PR #112, opened 2017-09-11
[HIVEMALL-133][SPARK] Support spark-v2.2 in the hivemalls-spark module

maropu:
This pr added supports for spark-2.2 in Hivemall.

This pr is currently WIP because:
1. Java7's been dropped in spark-v2.2 and Hivemall still supports it, so we need some entries to check a Java version when `spark-2.2` enabled.
2. We need to move common code into `spark/spark-common`.

## What type of PR is it?
Improvement

## What is the Jira issue?
https://issues.apache.org/jira/browse/HIVEMALL-133

## How was this patch tested?
Existing tests


myui:
👍  Merged!

PR #118, opened 2017-10-02
[HIVEMALL-146] Yet another UDF to generate n-grams

takuti:

Add a new UDF `to_ngrams(array<string> words, int minSize, int maxSize)` which returns list of n-grams `minSize <= n <= maxSize` for given words. This UDF can be alternative of the original Hive `ngrams` function.

## What type of PR is it?

Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-146

## How was this patch tested?

Unit test, manual tests both on EMR and local Hive

## How to use this feature?

as documented

## Checklist

(Please remove this section if not needed; check `x` for YES, blank for NO)

### DIFF
--- a/core/src/main/java/hivemall/tools/text/NgramsUDF.java
+++ b/core/src/main/java/hivemall/tools/text/NgramsUDF.java
@@ -29,6 +29,7 @@ import org.apache.hadoop.io.Text;
 
 import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -38,7 +39,8 @@ import java.util.List;
 @UDFType(deterministic = true, stateful = false)
 public final class NgramsUDF extends UDF {
 
-    public List<Text> evaluate(final List<Text> words, final int minSize, final int maxSize)
+    @Nullable
+    public List<Text> evaluate(@Nullable final List<Text> words, final int minSize, final int maxSize)
             throws HiveException {
         if (words == null) {
             return null;
--- a/core/src/main/java/hivemall/tools/text/NgramsUDF.java
+++ b/core/src/main/java/hivemall/tools/text/NgramsUDF.java
@@ -18,8 +18,6 @@
  */
 package hivemall.tools.text;
 
-import hivemall.utils.lang.StringUtils;
-
 import org.apache.hadoop.hive.ql.exec.Description;
 import org.apache.hadoop.hive.ql.exec.UDF;
 import org.apache.hadoop.hive.ql.exec.UDFArgumentException;
@@ -40,8 +38,8 @@ import java.util.List;
 public final class NgramsUDF extends UDF {
 
     @Nullable
-    public List<Text> evaluate(@Nullable final List<Text> words, final int minSize, final int maxSize)
-            throws HiveException {
+    public List<Text> evaluate(@Nullable final List<Text> words, final int minSize,
+            final int maxSize) throws HiveException {
         if (words == null) {
             return null;
         }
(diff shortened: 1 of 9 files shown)


core/src/main/java/hivemall/tools/text/NgramsUDF.java
```diff
@@ -0,0 +1,76 @@
...
+            for (int ngramSize = minSize; ngramSize <= maxSize; ngramSize++) {
+                if (i + ngramSize > numWords) { // exceeds the final element
+                    continue;
+                }
+
+                final List<String> ngram = new ArrayList<String>();
+                for (int j = i; j < i + ngramSize; j++) {
+                    ngram.add(words.get(j).toString());
+                }
+                ngrams.add(new Text(StringUtils.join(ngram, " ")));
```
myui: empty string is added where `i + ngramSize == numWords`?

takuti: `ngram` will never be empty thanks to `if (i + ngramSize > numWords) continue;`. Notice that the for-loop creating `ngram` increments `j` from `i` to `i + ngramSize - 1`.

core/src/main/java/hivemall/tools/text/NgramsUDF.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.hadoop.hive.ql.udf.UDFType;
+import org.apache.hadoop.io.Text;
+
+import javax.annotation.Nonnegative;
+import javax.annotation.Nonnull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Description(name = "to_ngrams", value = "_FUNC_(array<string> words, int minSize, int maxSize])"
```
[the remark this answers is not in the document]
takuti: I agree, but, at the same time, I like to use intuitive and well-known term as a name of such simple function. Let me think.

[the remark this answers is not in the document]
takuti: Since the function is applicable for a list of characters (character-based ngrams), using a word `word` e.g., `wordgram` sounds inappropriate for me. 

How about:

- `ngrams_joined_all` 
  - in contrast to the original `ngrams` which returns top-k most-frequent ngrams which contains "ngram" represented as a list of words
- `ngrams_between`, `ngrams_in_range` 
  - our UDF returns ngrams where `min <= n <= max`
- `to_ngram_list` 
  - a little bit longer than `to_ngrams` and less likely to be confusing 
  - `ngrams` creates list of named structs, but ours simply returns list of ngrams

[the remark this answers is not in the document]
myui: I user want character-level ngram, existing ngram is fine.

This function is for word-level ngram; and thus wordgram seems fine.

core/src/main/java/hivemall/tools/text/NgramsUDF.java
myui: I see. Then, avoid using ArrayList as follows and consider null in words array.

```java
final int numWords = words.size();
for (int i = 0; i < numWords; i++) {
 for (int ngramSize = minSize; ngramSize <= maxSize; ngramSize++) {
   final int end = i+ngram;
   if(end > numWords) {
      continue;
   }
   final StringBuilder buf = new StringBuilder();
   for(int j=i; j<end; j++) {
      Text w = words.get(i);
      if(w == null) { continue; } // avoid "null" to be added or throw exception
      if(j!=i) { buf.append(' '));
      buf.append(w.toString()).append(' ');
   }
   ngrams.add(new Text(buf.toString())));
```

core/src/main/java/hivemall/tools/text/NgramsUDF.java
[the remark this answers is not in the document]
takuti: + I've noticed that, since the results are joined with whitespace, using this new UDF for character-based ngram is infeasible.

Let's say `word_ngrams`.

core/src/main/java/hivemall/tools/text/NgramsUDF.java
takuti: Good point - `StringUtils.join()` adds a whitespace even to NULL element.

core/src/main/java/hivemall/tools/text/NgramsUDF.java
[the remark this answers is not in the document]
myui: Sounds good.

core/src/main/java/hivemall/tools/text/WordNgramsUDF.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new UDFArgumentException("`maxSize` must be greater than or equal to `minSize`: "
+                    + maxSize);
+        }
+        return getNgrams(words, minSize, maxSize);
+    }
+
+    @Nonnull
+    private static List<Text> getNgrams(@Nonnull final List<Text> words,
+            @Nonnegative final int minSize, @Nonnegative final int maxSize) throws HiveException {
+        final List<Text> ngrams = new ArrayList<Text>();
```
myui: final StringBuilder ngram = new StringBuilder();

core/src/main/java/hivemall/tools/text/WordNgramsUDF.java
```diff
@@ -0,0 +1,85 @@
...
+                        throw new HiveException(
+                            "`array<string> words` must not contain NULL element");
+                    }
+                    if (j > i) { // insert single whitespace between elements
+                        ngram.append(" ");
+                    }
+                    ngram.append(word.toString());
+                }
+                ngrams.add(new Text(ngram.toString()));
+            }
```
myui: StringUtils.clear(ngram);

myui: @takuti oops... Could you reuse StringBuilder instance as commented?

core/src/main/java/hivemall/tools/text/WordNgramsUDF.java
```diff
@@ -0,0 +1,85 @@
...
+                final int end = i + ngramSize;
+                if (end > numWords) { // exceeds the final element
+                    continue;
+                }
+
+                final StringBuilder ngram = new StringBuilder();
+                for (int j = i; j < end; j++) {
+                    final Text word = words.get(j);
+                    if (word == null) {
+                        throw new HiveException(
```
myui: UDFArgumentException is more appropriate.

core/src/main/java/hivemall/tools/text/WordNgramsUDF.java
takuti: 👍

myui:
@takuti LGTM. Can you merge this PR with squashing.

myui:
I personally prefer `wordgrams` though.
http://search.cpan.org/dist/Text-WordGrams/lib/Text/WordGrams.pm

PR #122, opened 2017-10-15
[HIVEMALL-147][Spark] Support all Hivemall functions of v0.5-rc.1 in Spark Dataframe

maropu:
This pr added more Hivemall functions for Spark DataFrame. However, some of the functions are not supported here because Spark simply cannot handle them (e.g., unsupported types, returned types depending on options, ...).

## What type of PR is it?
Feature

## What is the Jira issue?
https://issues.apache.org/jira/browse/HIVEMALL-147

## How was this patch tested?
Added tests in `HivemallOpsWithFeatureSuite`.

### DIFF
--- a/core/src/main/java/hivemall/evaluation/AUCUDAF.java
+++ b/core/src/main/java/hivemall/evaluation/AUCUDAF.java
@@ -110,7 +110,7 @@ public final class AUCUDAF extends AbstractGenericUDAFResolver {
 
         @Override
         public ObjectInspector init(Mode mode, ObjectInspector[] parameters) throws HiveException {
-            assert (parameters.length == 2 || parameters.length == 3) : parameters.length;
+            assert (0 < parameters.length && parameters.length <= 3) : parameters.length;
             super.init(mode, parameters);
 
             // initialize input
@@ -439,7 +439,7 @@ public final class AUCUDAF extends AbstractGenericUDAFResolver {
 
         @Override
         public ObjectInspector init(Mode mode, ObjectInspector[] parameters) throws HiveException {
-            assert (parameters.length == 2 || parameters.length == 3) : parameters.length;
+            assert (0 < parameters.length && parameters.length <= 3) : parameters.length;
             super.init(mode, parameters);
 
             // initialize input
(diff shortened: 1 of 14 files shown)


core/src/main/java/hivemall/evaluation/AUCUDAF.java
```diff
@@ -110,7 +110,7 @@ public ClassificationEvaluator() {}
 
         @Override
         public ObjectInspector init(Mode mode, ObjectInspector[] parameters) throws HiveException {
-            assert (parameters.length == 2 || parameters.length == 3) : parameters.length;
+            assert (0 < parameters.length && parameters.length <= 3) : parameters.length;
```
maropu: In Spark, this assertion fails because Spark passes a single parameter in `parameters` here for final output (IIUC [`AUC` finally outputs a single double-typed value for each group](https://github.com/apache/incubator-hivemall/pull/122/files#diff-9d758588c8fad559a15d0b2362e757b2R1134)). In Hive, does this work well?

core/src/main/java/hivemall/evaluation/HitRateUDAF.java
```diff
@@ -71,9 +71,6 @@
                 + " - Returns HitRate")
 public final class HitRateUDAF extends AbstractGenericUDAFResolver {
 
-    // prevent instantiation
-    private HitRateUDAF() {}
-
```
maropu: This prevents Spark from loading UDAFs by using reflection. Can we remove this?

core/src/main/java/hivemall/evaluation/AUCUDAF.java
myui: It would be a bug and should use Preconditions. I guess enable assertion `-ea`  is not enabled (cc: @takuti )

core/src/main/java/hivemall/evaluation/HitRateUDAF.java
myui: sure.

myui:
LGTM. Merged! Thanks.

PR #124, opened 2017-10-30
[HIVEMALL-157] Avoid Null Pointer Exception caused by uninitialized queue handler

takuti:

Even though `to_ordered_list` allows (and ignores) NULL inputs, following query fails due to NPE:

```sql
select to_ordered_list(null, null)
```
> Null Pointer Exception

This PR fixes the problem; now, the function returns empty list in case that queue handler is uninitialized because of NULL inputs:

> []

## What type of PR is it?

Bug Fix

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-157

## How was this patch tested?

Manually tested on local and EMR Hive

## Checklist

### DIFF
--- a/core/src/main/java/hivemall/tools/list/UDAFToOrderedList.java
+++ b/core/src/main/java/hivemall/tools/list/UDAFToOrderedList.java
@@ -311,11 +311,11 @@ public final class UDAFToOrderedList extends AbstractGenericUDAFResolver {
             QueueAggregationBuffer myagg = (QueueAggregationBuffer) agg;
 
             Pair<List<Object>, List<Object>> tuples = myagg.drainQueue();
-            List<Object> keyList = tuples.getKey();
-            List<Object> valueList = tuples.getValue();
-            if (valueList.isEmpty()) {
+            if (tuples == null) {
                 return null;
             }
+            List<Object> keyList = tuples.getKey();
+            List<Object> valueList = tuples.getValue();
 
             Object[] partialResult = new Object[4];
             partialResult[0] = valueList;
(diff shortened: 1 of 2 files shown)


core/src/main/java/hivemall/tools/list/UDAFToOrderedList.java
```diff
@@ -406,6 +406,11 @@ void merge(@Nonnull List<Object> o_keyList, @Nonnull List<Object> o_valueList) {
 
             @Nonnull
             Pair<List<Object>, List<Object>> drainQueue() {
+                if (queueHandler == null) {
+                    return Pair.of((List<Object>) new ArrayList<Object>(),
+                        (List<Object>) new ArrayList<Object>());
```
myui: better to return `null` following the way in other SQL standard functions.

`terminate()`, `terminatePartial()`, and `@Nonnull` annotation in `drainQueue` also need to be changed.

takuti: @myui Updated. PTAL!

core/src/main/java/hivemall/tools/list/UDAFToOrderedList.java
```diff
@@ -404,11 +407,9 @@ void merge(@Nonnull List<Object> o_keyList, @Nonnull List<Object> o_valueList) {
                 }
             }
 
-            @Nonnull
             Pair<List<Object>, List<Object>> drainQueue() {
```
myui: Add `@Nullable` for the return of `drainQueue` .

core/src/main/java/hivemall/tools/list/UDAFToOrderedList.java
```diff
@@ -363,6 +363,9 @@ public void merge(@SuppressWarnings("deprecation") AggregationBuffer agg, Object
                 throws HiveException {
             QueueAggregationBuffer myagg = (QueueAggregationBuffer) agg;
             Pair<List<Object>, List<Object>> tuples = myagg.drainQueue();
+            if (tuples == null) {
+                return new ArrayList<Object>();
```
myui: should be `return null;`.

myui:
@takuti LGTM. You can merge it with squashing.

PR #143, opened 2018-04-13
[HIVEMALL-189] Create a list of all functions

takuti:

Create a list of all functions in the documentation. In order to make maintenance easier and simpler, the list is systematically generated by reading `@Description` annotation in the code: [takuti/hivemalldoc](https://github.com/takuti/hivemalldoc). 

In case this list does not look sufficient, let's update `@Description` annotation itself and make the code more informative in the future.

## What type of PR is it?

Documentation

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-189

### DIFF
--- a/docs/gitbook/SUMMARY.md
+++ b/docs/gitbook/SUMMARY.md
@@ -26,31 +26,31 @@
     * [Install as permanent functions](getting_started/permanent-functions.md)
     * [Input Format](getting_started/input-format.md)
 
-* [List of functions](misc/funcs.md)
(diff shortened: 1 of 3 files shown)


myui:
@takuti I think it's better to have hivemalldoc module (e.g., `tools/hivemalldoc` )

Hadoop tools.
https://github.com/apache/hadoop/tree/trunk/hadoop-tools

Here is tips to include Gradle module. 
https://stackoverflow.com/questions/40352951/include-gradle-module-to-maven-project

PR #145, opened 2018-04-16
[HIVEMALL-191] Add Kryo serialization test to existing workaround code

takuti:

Add Kryo serialization test to existing workaround code as: https://github.com/apache/incubator-hivemall/commit/f6765dff7be67e1a3327709bbb9bfdc6eba7b97f

To be more precise, currently two UDFs `quantified_features` and `tokenize_ja` explicitly have the workaround lazy instantiation code. So, this PR makes their `transient` keyword unnecessary.

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-191

## How was this patch tested?

Added some unit tests, and manually tested as well

## Checklist

### DIFF
--- a/core/src/main/java/hivemall/ftvec/trans/QuantifiedFeaturesUDTF.java
+++ b/core/src/main/java/hivemall/ftvec/trans/QuantifiedFeaturesUDTF.java
@@ -48,9 +48,6 @@ public final class QuantifiedFeaturesUDTF extends GenericUDTF {
     private Identifier<String>[] identifiers;
     private DoubleWritable[] columnValues;
 
-    // lazy instantiation to avoid org.apache.hive.com.esotericsoftware.kryo.KryoException: java.lang.NullPointerException
-    private transient Object[] forwardObjs;
-
     @SuppressWarnings("unchecked")
     @Override
     public StructObjectInspector initialize(ObjectInspector[] argOIs) throws UDFArgumentException {
@@ -65,7 +62,6 @@ public final class QuantifiedFeaturesUDTF extends GenericUDTF {
         this.doubleOIs = new PrimitiveObjectInspector[outputSize];
         this.columnValues = new DoubleWritable[outputSize];
         this.identifiers = new Identifier[outputSize];
-        this.forwardObjs = null;
 
         for (int i = 0; i < outputSize; i++) {
             columnValues[i] = new DoubleWritable(Double.NaN);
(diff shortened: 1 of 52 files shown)


takuti:
I just `grep KryoException` to find target UDFs. @myui Please let me know if there are other UDFs we have to test/fix. Maybe, should we add `testSerialization` to all existing UDF tests?

takuti:
// Fixing scala test failure

myui:
@takuti Ideally, we SHOULD have serialization tests for all UDFs. New UDFs MUST have a serialization test.

takuti:
Added serialization test case to all existing GenericUDF/UDTF tests. Vanilla UDF and UDAF tests are not changed.

@myui Could you review?

myui:
Sure.

core/src/main/java/hivemall/ftvec/trans/QuantifiedFeaturesUDTF.java
```diff
@@ -87,30 +80,27 @@ public StructObjectInspector initialize(ObjectInspector[] argOIs) throws UDFArgu
...
                     } else {
                         String k = arg.toString();
                         int id = identifier.valueOf(k);
-                        values[i].set(id);
+                        values[i] = id;
                     }
                 }
             }
-            forward(forwardObjs);
+            forward(new Object[] {WritableUtils.toWritableList(values)});
```
myui: Why you made this change? Creating object (calling `WritableUtils.toWritableList`) for each `forward` is costful.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
```diff
@@ -69,13 +69,10 @@
     private static final int READ_TIMEOUT_MS = 60000; // 60 sec
     private static final long MAX_INPUT_STREAM_SIZE = 32L * 1024L * 1024L; // ~32MB
 
-    private Mode _mode;
```
myui: Any serialization problems in `_mode` and `_stopTagsArray` ?
Any problem in those early instantiation?

takuti: As far as we can observe from the result of test case `KuromojiUDFTest#testSerialization`, there is no serialization issue here.

What this modification does is to defensively avoid early instantiation of Lucene's unclear classes (`Mode`, `CharArraySet` and `UserDictionary`).

myui:
Merged. Thanks. I'll do some refactoring in the following commit (TestUtils.java is redundant etc).

```
TestUtils.java
TestBinariseLabelUDTF.java
QuantifiedFeaturesUDTF.java
KuromojiUDF.java
```

core/src/main/java/hivemall/ftvec/trans/QuantifiedFeaturesUDTF.java
```diff
@@ -87,32 +87,37 @@ public StructObjectInspector initialize(ObjectInspector[] argOIs) throws UDFArgu
...
     @Override
     public void process(Object[] args) throws HiveException {
-        if (forwardObjs == null) {
-            this.forwardObjs = new Object[] {Arrays.asList(columnValues)};
-        }
-
+        int outputSize = args.length - 1;
         boolean outputRow = boolOI.get(args[0]);
         if (outputRow) {
-            final DoubleWritable[] values = this.columnValues;
```
myui: @takuti local variable access is more faster than instance variable.

core/src/main/java/hivemall/ftvec/trans/QuantifiedFeaturesUDTF.java
```diff
@@ -87,32 +87,37 @@ public StructObjectInspector initialize(ObjectInspector[] argOIs) throws UDFArgu
...
-            for (int i = 0, outputSize = args.length - 1; i < outputSize; i++) {
+            if (forwardObjs == null) {
+                // forwardObjs internally references columnValues
+                List<DoubleWritable> column = new ArrayList<>(outputSize);
+                this.forwardObjs = new Object[] {column};
```
myui: why not just `new Object[] {Arrays.asList(columnValues)};` as old code.

nlp/src/main/java/hivemall/nlp/tokenizer/KuromojiUDF.java
myui: Mode is enum and serializable and thus better to remain the old code.

PR #167, opened 2018-10-18
[HIVEMALL-220] Implement Cofactor

jaxony:

Implemented new matrix factorization algorithm for the recommendation problem.

## What type of PR is it?

Feature

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-220

## How was this patch tested?

Unit tests and manual testing on ML20M in a Hive dev environment

## How to use this feature?

TODO

## Checklist

(Please remove this section if not needed; check `x` for YES, blank for NO)

- [ ] Did you apply source code formatter, i.e., `./bin/format_code.sh`, for your commit?
- [ ] Did you run system tests on Hive (or Spark)?

### DIFF
--- a/core/src/main/java/hivemall/fm/Feature.java
+++ b/core/src/main/java/hivemall/fm/Feature.java
@@ -171,7 +171,7 @@ public abstract class Feature {
     }
 
     @Nonnull
-    static Feature parseFeature(@Nonnull final String fv, final boolean asIntFeature)
+    public static Feature parseFeature(@Nonnull final String fv, final boolean asIntFeature)
             throws HiveException {
         final int pos1 = fv.indexOf(':');
         if (pos1 == -1) {
--- a/core/src/main/java/hivemall/mf/CofactorModel.java
+++ b/core/src/main/java/hivemall/mf/CofactorModel.java
@@ -264,7 +264,8 @@ public class CofactorModel {
 
     }
 
-    private static RealMatrix calculateDelta(List<Feature> children, Map<String, RealVector> weights, int numFactors, float constant) {
+    protected static RealMatrix calculateDelta(List<Feature> children, Map<String, RealVector> weights, int numFactors, float constant) {
+        // equivalent to `B_u.T.dot((c1 - c0) * B_u)` in cofacto.py
         RealMatrix delta = new Array2DRowRealMatrix(numFactors, numFactors);
         int i = 0, j = 0;
         for (int f = 0; f < numFactors; f++) {
@@ -279,9 +280,8 @@ public class CofactorModel {
     private static double dotFactorsAlongDims(List<Feature> keys, Map<String, RealVector> weights, int dim1, int dim2) {
         double result = 0.d;
         for (Feature f : keys) {
-            double rating = f.getValue();
             RealVector vec = getFactorVector(f.getFeature(), weights);
-            result += rating * vec.getEntry(dim1) * vec.getEntry(dim2);
+            result += vec.getEntry(dim1) * vec.getEntry(dim2);
         }
         return result;
     }
--- a/core/src/main/java/hivemall/mf/CofactorModel.java
+++ b/core/src/main/java/hivemall/mf/CofactorModel.java
@@ -219,7 +219,7 @@ public class CofactorModel {
 
         for (CofactorizationUDTF.TrainingSample sample : samples) {
             // filter for trainable items
-            List<Feature> trainableItems = filterTrainableFeatures(sample, beta);
+            List<Feature> trainableItems = filterTrainableFeatures(sample.children, beta);
 
             RealVector A = calculateA(trainableItems, beta, c1);
 
@@ -257,10 +257,10 @@ public class CofactorModel {
         return idMatrix.scalarMultiply(lambda);
     }
 
-    private static List<Feature> filterTrainableFeatures(CofactorizationUDTF.TrainingSample sample, Map<String, RealVector> weights) {
+    private static List<Feature> filterTrainableFeatures(Feature[] features, Map<String, RealVector> weights) {
         List<Feature> trainableFeatures = new ArrayList<>();
         String fName;
-        for (Feature f : sample.children) {
+        for (Feature f : features) {
             fName = f.getFeature();
             if (isTrainable(fName, weights)) {
                 trainableFeatures.add(f);
--- a/core/src/main/java/hivemall/mf/CofactorModel.java
+++ b/core/src/main/java/hivemall/mf/CofactorModel.java
@@ -148,7 +148,7 @@ public class CofactorModel {
         String key = f.getFeature();
         biases.put(key, value);
     }
-    
+
     public RealVector getGammaVector(final Feature f) {
         return getFactorVector(f, gamma,false);
     }
@@ -163,15 +163,6 @@ public class CofactorModel {
 
     public void setGammaBias(final Feature f, final double value) {
         setBias(f, gammaBias, value);
-    }
-
-    @Nonnull
-    private static Random[] newRandoms(@Nonnull final int size, final long seed) {
-        final Random[] rand = new Random[size];
-        for (int i = 0, len = rand.length; i < len; i++) {
-            rand[i] = new Random(seed + i);
-        }
-        return rand;
     }
 
     public double getMeanRating() {
@@ -198,6 +189,14 @@ public class CofactorModel {
         return getFactorVector(f, beta, init);
     }
 
+    public double getBetaBias(final Feature f) {
+        return getBias(f, betaBias);
+    }
+
+    public void setBetaBias(final Feature f, final double value) {
+        setBias(f, betaBias, value);
+    }
+
     /**
      * Update latent factors of the users in the provided mini-batch.
      */
@@ -205,12 +204,13 @@ public class CofactorModel {
 
     }
 
-    public double getBetaBias(final Feature f) {
-        return getBias(f, betaBias);
-    }
-
-    public void setBetaBias(final Feature f, final double value) {
-        setBias(f, betaBias, value);
+    @Nonnull
+    private static Random[] newRandoms(@Nonnull final int size, final long seed) {
+        final Random[] rand = new Random[size];
+        for (int i = 0, len = rand.length; i < len; i++) {
+            rand[i] = new Random(seed + i);
+        }
+        return rand;
     }
 
     protected static void uniformFill(final RealVector a, final Random rand, final float maxInitValue) {
--- a/core/src/main/java/hivemall/mf/CofactorModel.java
+++ b/core/src/main/java/hivemall/mf/CofactorModel.java
@@ -234,7 +234,7 @@ public class CofactorModel {
 
         RealVector A = calculateA(trainableItems, beta, c1);
 
-        RealMatrix delta = calculateDelta(trainableItems, beta, numFactors, c1 - c0);
+        RealMatrix delta = calculateWTWSubset(trainableItems, beta, numFactors, c1 - c0);
         RealMatrix B = BTBpR.add(delta);
 
         // solve and update factors
@@ -271,8 +271,8 @@ public class CofactorModel {
         RealVector RSD = calculateRSD(sample.parent, trainableCooccurringItems, numFactors, betaBias, gammaBias, gamma);
         RealVector ApRSD = calculateA(trainableUsers, theta, c1).add(RSD);
 
-        RealMatrix GTG = calculateDelta(trainableCooccurringItems, gamma, numFactors, 1.f);
-        RealMatrix delta = calculateDelta(trainableUsers, theta, numFactors, c1 - c0);
+        RealMatrix GTG = calculateWTWSubset(trainableCooccurringItems, gamma, numFactors, 1.f);
+        RealMatrix delta = calculateWTWSubset(trainableUsers, theta, numFactors, c1 - c0);
         RealMatrix B = TTTpR.add(delta).add(GTG);
 
         // solve and update factors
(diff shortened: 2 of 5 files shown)


core/src/main/java/hivemall/mf/FactorizedModel.java
```diff
@@ -30,25 +30,25 @@
 import javax.annotation.concurrent.NotThreadSafe;
 
 @NotThreadSafe
-public final class FactorizedModel {
+public class FactorizedModel {
```
myui: It seems FactorizedModel is not used in Cofactor. 

Is this change required? Revert if not used.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+    // rank matrix initialization
+    protected final RankInitScheme initScheme;
+
+    @Nonnull
+    private double globalBias;
+
+    // storing trainable latent factors and weights
+    private Map<String, RealVector> theta;
+    private Map<String, RealVector> beta;
+    private Map<String, Double> betaBias;
```
myui: Please use `Object2DoubleMap<String> betaBias` instead to reduce memory consumption.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+    private Map<String, RealVector> gamma;
+    private Map<String, Double> gammaBias;
```
myui: Please use `Object2DoubleMap<String> gammaBias` instead to reduce memory consumption.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+                WTW.setEntry(f, ff, val);
+            }
+        }
+        return WTW;
+    }
+
+    protected static RealMatrix calculateWTWSubset(List<Feature> subset, Map<String, RealVector> weights, int numFactors, float constant) {
+        // equivalent to `B_u.T.dot((c1 - c0) * B_u)` in cofacto.py
+        RealMatrix delta = new Array2DRowRealMatrix(numFactors, numFactors);
+        int i = 0, j = 0;
```
myui: ` int i = 0, j = 0;` never used. 

This could be found in IDE warning. Please resolve all IDE warnings.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+        this.c0 = c0;
+        this.c1 = c1;
+
+    }
+
+    private void initFactorVector(String key, Map<String, RealVector> weights) {
+        if (weights.containsKey(key)) {
+            return;
+        }
+        RealVector v = new ArrayRealVector(factor);
```
myui:
```
final double[] v = new double[factor];
...
weight.put(key, new ArrayRealVector(v));
```

is better.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+
+    protected static void gaussianFill(final RealVector a, final Random[] rand, final double stddev) {
+        for (int i = 0, len = a.getDimension(); i < len; i++) {
+            double v = MathUtils.gaussian(0.d, stddev, rand[i]);
+            a.setEntry(i, v);
+        }
+    }
+
+    private static void checkHyperparameterC(final float c) {
+        assert c >= 0.f && c <= 1.f;
```
myui: Better to throw exceptions instead of assertion (could be off in production).

`Precondition.checkArgument(c >= 0.f && c <= 1.f)`

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+        for (Feature f : keys) {
+            RealVector vec = getFactorVector(f.getFeature(), weights);
+            result += vec.getEntry(dim1) * vec.getEntry(dim2);
+        }
+        return result;
+    }
+
+    private static double dotFactorsAlongDims(Map<String, RealVector> weights, int dim1, int dim2) {
+        double result = 0.d;
+        for (String key : weights.keySet()) {
```
myui: Why not simply 

```
        for(RealVector vec : weights.values()) {
           ...      
        }
```

What's expected behavior when `vec.getEntry(dim1/2)` returns null?

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,629 @@
...
+
+    /**
+     * Add v to u in-place without creating a new RealVector instance.
+     * @param u vector to which v will be added
+     * @param v vector containing new values to be added to u
+     * @param scalar value to multiply each entry in v before adding to u
+     */
+    private static void addInPlace(RealVector u, RealVector v, double scalar) {
+        assert u.getDimension() == v.getDimension();
+        for (int i = 0; i < u.getDimension(); i++) {
```
myui: I prefer avoiding method calls by

```java
for (int i = 0, size=u.getDimension(); i < size; i++) {
```

core/src/main/java/hivemall/mf/CofactorizationUDTF.java
```diff
@@ -0,0 +1,574 @@
...
+        private List<TrainingSample> users;
+        private List<TrainingSample> items;
+
+        protected MiniBatch(int maxSize) {
+            this.maxSize = maxSize;
+        }
+
+        protected void add(TrainingSample sample) {
+            if (size() == this.maxSize) {
+                return;
```
myui: `sample` is thrown away?

You can return boolean value to check MiniBatch is full or not.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,638 @@
...
+        return weights.get(key);
+    }
+
+    private static void setFactorVector(String key, Map<String, RealVector> weights, RealVector factorVector) {
+        assert weights.containsKey(key);
+        weights.put(key, factorVector);
+    }
+
+    private static double getBias(String key, Map<String, Double> biases) {
+        if (!biases.containsKey(key)) {
```
myui: three hash lookup for worse case...

```
        final Double v = biases.get(key);
        if(v == null) {
            return 0.d;
        }
        return v.doubleValue();
```

Or, 

```java
    private static double getBias(String key, Object2DoubleMap<String> biases) {
        // biases.defaultReturnValue(0.f); -- set in initialization
        return biases.get(key);
    }
```

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,640 @@
...
+
+        for (CofactorizationUDTF.TrainingSample sample : samples) {
+            RealVector newThetaVec = calculateNewThetaVector(sample, beta, factor, BTBpR, c0, c1);
+            if (newThetaVec != null) {
+                setFactorVector(sample.context.getFeature(), theta, newThetaVec);
+            }
+        }
+    }
+
+    protected static RealVector calculateNewThetaVector(CofactorizationUDTF.TrainingSample sample, Map<String, RealVector> beta,
```
myui: `@VisibleForTesting` for testing purpose.

core/src/main/java/hivemall/mf/CofactorizationUDTF.java
jaxony: I think an explicit mini batch size is not required anymore. The upper limit is the size of the `inputBuf`. Is this correct?

core/src/main/java/hivemall/mf/CofactorModel.java
jaxony: `RealVector` implementation has now changed to `double[]`, so by default it should be `0`. In any case, all entries in `weights` have been initialized to uniform random or gaussian so the array entries should be valid.

core/src/main/java/hivemall/fm/Feature.java
```diff
@@ -383,4 +383,10 @@ public static void l2normalize(@Nonnull final Feature[] features) {
         }
     }
 
+    @Override
```
myui: Why this `equals` method is required? Assume this is not used.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+
+public class CofactorModel {
+
+    public enum RankInitScheme {
+        random /* default */, gaussian;
+
```
myui: please remove unnessesary line breaks.

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+        this.A = new ArrayRealVector(this.factor);
+
+        this.randU = newRandoms(factor, 31L);
+        this.randI = newRandoms(factor, 41L);
+
+        Preconditions.checkArgument(c0 >= 0.f && c0 <= 1.f);
+        Preconditions.checkArgument(c1 >= 0.f && c1 <= 1.f);
+        this.c0 = c0;
+        this.c1 = c1;
+
```
myui: break line

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+    protected static RealVector solve(RealMatrix B, double[][] dataB, RealVector A, double[] dataA) throws HiveException {
+        // b * x = a
+        // solves for x
+        copyData(B, dataB);
+        copyData(A, dataA);
+        SingularValueDecomposition svd = new SingularValueDecomposition(B);
+        return svd.getSolver().solve(A);
+    }
+
+    private static void copyData(RealMatrix dst, double[][] src) throws HiveException {
```
myui: add `@Nonnull`

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+    }
+
+    private static void copyData(RealVector dst, double[] src) throws HiveException {
+        checkCondition(dst.getDimension() == src.length, DIFFERENT_DIMS_ERR);
+        for (int i = 0; i < dst.getDimension(); i++) {
+            dst.setEntry(i, src[i]);
+        }
+    }
+
+    private static void copyData(double[] dst, RealVector src) throws HiveException {
```
myui: add `@Nonnull` .. ditto

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+        return result;
+    }
+
+    @VisibleForTesting
+    protected static double[] calculateA(List<Feature> items, Map<String, double[]> weights, int numFactors, float constant) throws HiveException {
+        // Equivalent to: a = x_u.dot(c1 * B_u)
+        // x_u is a (1, i) matrix of all ones
+        // B_u is a (i, F) matrix
+        // What it does: sums factor n of each item in B_u
+//        clearArray(A);
```
myui: remove unused code

core/src/main/java/hivemall/mf/CofactorModel.java
```diff
@@ -0,0 +1,715 @@
...
+    }
+
+    private static void clearArray(double[] a) {
+        for (int i = 0; i < a.length; i++) {
+            a[i] = 0.d;
+        }
+    }
+
+    public Double predict(String user, String item) {
+        if (!isTrainable(user, theta) || !isTrainable(item, beta)) {
```
myui: naming can be revised. `isTrainable` in predict? 

Rename or simply removing method. `!theta.contains(user) || beta.contains(item)`

core/src/main/java/hivemall/fm/Feature.java
jaxony: I use it in a test in `CofactorizationUDTF` where I check if a `Feature` read from disk is equal to the expected `Feature`

core/src/main/java/hivemall/mf/CofactorizationUDTF.java
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+    @Nonnull
+    private static Feature instantiateFeature(@Nonnull final ByteBuffer input) {
+        return new StringFeature(input);
+    }
+
+    @VisibleForTesting
+    protected boolean readMiniBatchFromFile(MiniBatch miniBatch) throws HiveException {
+        inputBuf.clear();
```
jaxony: Fix this bug: remove the line

core/src/main/java/hivemall/fm/Feature.java
myui: See https://medium.com/codelog/overriding-hashcode-method-effective-java-notes-723c1fedf51c 

Usually, overriding `equals` required `hashCode` because hashCode (and equals) is used for HashMap key search.

jaxony:
Hi @myui, I'd like to continue working on cofactor and hopefully get it merged in. Any recommendations on how I can test implementation in distributed environment?

myui:
Hi @jaxony welcome back!

For pseudo evaluation, docker is easy to evaluate.
https://hivemall.incubator.apache.org/userguide/docker/getting_started.html
^ single worker 

This docker-compose spin-up multiple hadoop servers using @Lewuathe 's
https://github.com/Lewuathe/docker-hadoop-cluster

Then, you can just login to master node and [setup Hive (and hivemall)](https://github.com/apache/incubator-hivemall/blob/master/resources/docker/Dockerfile#L58) on the master node. Note Hive command is just a standalone application submitting jobs to Hadoop.
https://github.com/apache/incubator-hivemall/blob/master/resources/docker/Dockerfile#L58

You can modify [Hive setup process on master node](https://github.com/Lewuathe/docker-hadoop-cluster/blob/master/hadoop-master/Dockerfile
) following 
https://github.com/apache/incubator-hivemall/blob/master/resources/docker/Dockerfile#L58

Also, I'll consider to provide you EMR testbed.

myui:
@jaxony For security reasons, we would like to add your IP address to whitelist for accessing our EMR environment. Do you have specific IP range (e.g., university's one)?

jaxony:
@myui I think the University of Melbourne's IP range is 128.250.0.0 to 128.250.0.255

PR #51, opened 2017-02-24
[HIVEMALL-75] Support Sparse Vector Format as the input of RandomForest

myui:
## What changes were proposed in this pull request?

Supported sparse vector as the input of RandomForest.

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-75

## How was this patch tested?

unit test and manual test

PR #178, opened 2019-01-18
[HIVEMALL-233] RandomForest regressor accepts sparse vector input

takuti:

Enable RandomForestRegressor to accept sparse vector input as RandomForestClassifier already does.

#51 made incomplete modifications on RandomForestRegressor, while its classifier counterpart has been properly updated.

## What type of PR is it?

Improvement

## What is the Jira issue?

https://issues.apache.org/jira/browse/HIVEMALL-233

## How was this patch tested?

Unit tests and manual test with tiny sample data:

```sql
with customers as (
  select 1 as id, "male" as gender, 23 as age, "Japan" as country, 12 as num_purchases
  union all
  select 2 as id, "female" as gender, 43 as age, "US" as country, 4 as num_purchases
  union all
  select 3 as id, "other" as gender, 19 as age, "UK" as country, 2 as num_purchases
  union all
  select 4 as id, "male" as gender, 31 as age, "US" as country, 20 as num_purchases
  union all
  select 5 as id, "female" as gender, 37 as age, "Australia" as country, 9 as num_purchases
),
training as (
  select
    array_concat(
      quantitative_features(
        array("age"),
        age
      ),
      categorical_features(
        array("country", "gender"),
        country, gender
      )
    ) as features,
    num_purchases
  from
    customers
)
select
  train_randomforest_regressor(
    feature_hashing(features), -- feature vector
    num_purchases, -- target value
    '-trees 40 -seed 31' -- hyper-parameters
  )
from
  training
;
```

## Checklist

- [x] Did you apply source code formatter, i.e., `./bin/format_code.sh`, for your commit?

### DIFF
--- a/core/src/main/java/hivemall/smile/classification/GradientTreeBoostingClassifierUDTF.java
+++ b/core/src/main/java/hivemall/smile/classification/GradientTreeBoostingClassifierUDTF.java
@@ -26,7 +26,10 @@ import hivemall.math.matrix.builders.RowMajorDenseMatrixBuilder;
 import hivemall.math.matrix.ints.ColumnMajorIntMatrix;
 import hivemall.math.random.PRNG;
 import hivemall.math.random.RandomNumberGeneratorFactory;
+import hivemall.math.vector.DenseVector;
+import hivemall.math.vector.SparseVector;
 import hivemall.math.vector.Vector;
+import hivemall.math.vector.VectorProcedure;
 import hivemall.smile.data.Attribute;
 import hivemall.smile.regression.RegressionTree;
 import hivemall.smile.utils.SmileExtUtils;
@@ -40,6 +43,8 @@ import hivemall.utils.math.MathUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
+import java.util.Map;
+import java.util.HashMap;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


core/src/main/java/hivemall/utils/collections/lists/DoubleArrayList.java
```diff
@@ -70,7 +70,7 @@ public DoubleArrayList add(@Nonnull double[] values) {
     private void expand(int max) {
         while (data.length < max) {
             final int len = data.length;
-            double[] newArray = new double[len * 2];
+            double[] newArray = new double[(len + 1) * 2];
```
myui: why this change has been made?

takuti: See d7695d461056b21eab25465e015c582edc2b57ce

In case # of samples is less than 10, [`SmileExtUtils#sort` calls `DoubleArrayList(0)`](https://github.com/apache/incubator-hivemall/blob/7b61a3be8178fef40154f4009fef2989a9fd4c49/core/src/main/java/hivemall/smile/utils/SmileExtUtils.java#L190-L192), and zero-sized array list falls into infinite loop in the `expand` method as `newArray = new double[0 * 2]`. The change prevents this.

myui: hmm, this PR does not resolves a potential bug in expand. Returning array should be `>= max` and `max` should be `minCapacity` where `expand`'s argument is expected to be >=1.

https://github.com/karussell/fastutil/blob/master/src/it/unimi/dsi/fastutil/doubles/DoubleArrayList.java#L203
https://github.com/karussell/fastutil/blob/master/src/it/unimi/dsi/fastutil/doubles/DoubleArrays.java#L136

Let me fix this in another PR.

PR #192, opened 2019-04-29
[HIVEMALL-253]UDF map_roulette() created

Solodye:
## What changes were proposed in this pull request?

Create UDF map_roulette(). 

## What type of PR is it?

Feature and documentation

## What is the Jira issue?

https://issues.apache.org/jira/projects/HIVEMALL/issues/HIVEMALL-253?filter=allopenissues
[HIVEMALL-253]

## How was this patch tested?

Unit test and handmade test:
```sql
-- 1. Empty map
select map_roulette(map()); -- NULL
-- 2. The map that key is null
select map_roulette(map(null, null, null, null)); -- NULL
-- 3. A map only one entry
select map_roulette(map(1, 2)); -- 1
-- 4. A map with normal value (1, 1) and null value (null, null)
select map_roulette(map(1, 2, null, null)); -- 50% 1, 50% null
-- 5. A map contains zero weight
select map_roulette(map(1, 2, 3, 0)); -- 1, but 3 may appear at a very low probability
-- 6. All weights are zero
select map_roulette(map(1, 0)); -- NULL
select map_roulette(map(1, 0, '5', 0)); -- NULL
-- 7. The testcase close to real data
select map_roulette(map(1, 0.5, 'Wang', 0.5)); -- 50% Wang, 50% 1
select map_roulette(map(1, 0.5, 'Wang', null)); -- 50% Wang, 50% 1
select map_roulette(map(1, 0.5, 'Wang', null, 'Zhang', null)); -- 1/3 Wang, 1/3 1, 1/3 Zhang
select map_roulette(map(1, null, 'Wang', null)); -- NULL
select map_roulette(map(1, 0, 'Wang', 0, 'Zhang', 1)); -- Zhang， but 'Wang', 'Zhang' may appear at a very low probability
select map_roulette(map('Wang', 3, 'Zhang', 2, 'Li', 6, 'Zhao', 8, 'Tom', 10)); -- 3/29 Wang, 2/29 Zhang, 6/29 Li, 8/29 Zhao, 10/29 Tom
-- 8. A usecase with wrong input
select map_roulette(map('Wong', 'A string', 'Zhao', 2));
-- Failed with exception java.io.IOException:org.apache.hadoop.hive.ql.metadata.HiveException: Error evaluating map_roulette([map('Wong':'A string','Zhao':2)])
-- Because weight must be a number.
-- 9. Negative weight
select map_roulette(map('Wong', 3, 'Zhao', -2));
[...]

PR #193, opened 2019-06-10
[HIVEMALL-253-2] map_roulette UDF

myui:
revise #192

### DIFF
--- a/core/src/main/java/hivemall/tools/map/MapRouletteUDF.java
+++ b/core/src/main/java/hivemall/tools/map/MapRouletteUDF.java
@@ -60,51 +60,6 @@ import org.apache.hadoop.hive.serde2.objectinspector.primitive.PrimitiveObjectIn
 @UDFType(deterministic = false, stateful = false) // it is false because it return value base on probability
 public final class MapRouletteUDF extends GenericUDF {
 
-    /**
-     * The map passed in saved all the value and its weight
-     *
-     * @param m A map contains a lot of item as key, with their weight as value
-     * @return The key that computer selected according to key's weight
-     */
-    @Nullable
-    private static Object algorithm(@Nonnull final Map<Object, Double> m) {
-        if (m.isEmpty()) {
-            return null;
-        }
-
-        // normalize the weight
-        double sum = 0;
-        for (Map.Entry<Object, Double> entry : m.entrySet()) {
-            sum += entry.getValue();
-        }
-        for (Map.Entry<Object, Double> entry : m.entrySet()) {
-            entry.setValue(entry.getValue() / sum);
-        }
-
-        // sort and generate a number axis
-        List<Map.Entry<Object, Double>> entryList = new ArrayList<>(m.entrySet());
-        Collections.sort(entryList, new KvComparator());
-        double tmp = 0;
-        for (Map.Entry<Object, Double> entry : entryList) {
-            tmp += entry.getValue();
-            entry.setValue(tmp);
-        }
-
-        // judge last value
-        if (entryList.get(entryList.size() - 1).getValue() > 1.0) {
-            entryList.get(entryList.size() - 1).setValue(1.0);
-        }
-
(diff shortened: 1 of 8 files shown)


Solodye:
OK, I'm watching this PR and learning from problems in my code.

core/src/main/java/hivemall/tools/map/MapRouletteUDF.java
```diff
@@ -49,164 +39,171 @@
...
-        if (m.size() == 0) {
+        final int size = m.size();
+        if (size == 0) {
             return null;
         }
-        // convert
-        Map<Object, Double> input = new HashMap<>();
-        Double avg = 0.0;
+
+        final Map<Object, Double> result = new HashMap<>(size);
```
Solodye: `Map<Object, Double> result` will be changed, why use final here?

myui: I cannot catch you. 

This instance is not changed and returned on `getObjectDoubleMap()`. The annotation means that `result` is not replaced.

`final` annotation for a map instance does not mean that `result` can not be updated.

Solodye: Got it, sorry for my poor English……

core/src/test/java/hivemall/tools/map/MapRouletteUDFTest.java
```diff
@@ -0,0 +1,153 @@
...
+                    new GenericUDF.DeferredObject[] {new GenericUDF.DeferredJavaObject(m)};
+            Object key = fmp.evaluate(arguments);
+            solve.put(key, solve.get(key) + 1);
+        }
+        List<Map.Entry<Object, Integer>> solveList = new ArrayList<>(solve.entrySet());
+        Collections.sort(solveList, new KvComparator());
+        Object highestSolve = solveList.get(solveList.size() - 1).getKey();
+        Assert.assertEquals(highestSolve.toString(), "Zhao");
+        Object secondarySolve = solveList.get(solveList.size() - 2).getKey();
+        Assert.assertEquals(secondarySolve.toString(), "Jerry");
```
myui: @Solodye 

`Tom` should be 2nd place in my understanding because average is `(0.2+0.1+0.1+0.5)/5=0.18`. 
Right? Then, why `Jerry` here? 

BTW, `Assert.assertEquals` takes `Assert.assertEquals(expected, actual)`, not `Assert.assertEquals(actual, expected)`.

myui: oops `(0.2+0.1+0.1+0.5)/4=0.225`. Ignore the comment the above.

myui: Since `(0.2+0.1+0.1+0.5)/5=0.18` in your code, `Jerry` is elected for the 2nd place.

Now, it's fixed to `(0.2+0.1+0.1+0.5)/4=0.225`, `Tom` should be elected for the 2nd place. Will fix the unit test.

core/src/main/java/hivemall/tools/map/MapRouletteUDF.java
```diff
@@ -0,0 +1,209 @@
...
+            Object key = entry.getKey();
+            result.put(key, Double.valueOf(v));
+
+            sum += v;
+            cnt++;
+        }
+
+        if (result.size() < m.size()) {
+            // fillna with the avg value
+            final Double avg = Double.valueOf(sum / cnt);
```
Solodye: Yes，`avg /= m.size();` was revised, so Jerry in testcase ' testRoulette()' should be elected for the 3rd place

myui:
merged. Thank you for your first contribution!

Solodye:
> merged. Thank you for your first contribution!

You are welcome, I have learned a lot from this contributing process and your work.

